PR #9, opened 2018-08-03
Support scalafixDependencies

### LINKED ISSUE #7: Add scalafixDependencies setting key to customise --tool-classpath
The scalafix cli supports an option --tool-classpath to classload custom rules from a custom classpath. Currently, sbt-scalafix does not make it possible to configure --tool-classpath. Here is a proposal to solve that

- Users specify custom rules as dependencies for the global setting key `scalafixDependencies: SettingKey[Seq[ModuleID]]`

```scala
scalafixDependencies += "org.scala-lang.scalafix" %% "2.13-migration" % "1.0.0"
```

- sbt-scalafix resolves the jars for that dependency and passes the jars to --tool-classpath

A commit by @MasseGuillaume implementing this feature is here https://github.com/MasseGuillaume/scalafix/commit/4108726c5239a584c440753fea9152b6d7467667

olafurpg:
Fixes #7.

This PR improves support for custom rules that are published to Maven
Central. This functionality is required for rules like
scala-collection-compat. We do a best-effort to avoid redundant
dependency fetching of `scalafixDependencies` by caching the results
in an in-memory map. In the future, it would make sense to expose more
powerful APIs in scalafix to cache classloaders between invocations.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -20,6 +20,9 @@ developers := List(
 resolvers += Resolver.sonatypeRepo("releases")
 libraryDependencies ++= List(
   "org.eclipse.jgit" % "org.eclipse.jgit" % "4.5.4.201711221230-r",
+  // coursier-small provides a binary stable API around Coursier making sure that
+  // sbt-scalafix doesn't conflict with the user's installed version of sbt-coursier.
+  // Details: https://github.com/olafurpg/coursier-small
   "com.geirsson" %% "coursier-small" % "1.0.0-M4",
   "org.scalatest" %% "scalatest" % "3.0.5" % Test
 )
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,7 @@ developers := List(
 resolvers += Resolver.sonatypeRepo("releases")
 libraryDependencies ++= List(
   "org.eclipse.jgit" % "org.eclipse.jgit" % "4.5.4.201711221230-r",
-  "com.geirsson" %% "coursier-small" % "1.0.0-M2",
+  "com.geirsson" %% "coursier-small" % "1.0.0-M4",
   "org.scalatest" %% "scalatest" % "3.0.5" % Test
 )
 
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -1,13 +1,19 @@
 package scalafix.sbt
 
-import sbt.File
+import com.geirsson.coursiersmall
+import com.geirsson.coursiersmall.CoursierSmall
+import com.geirsson.coursiersmall.Dependency
+import com.geirsson.coursiersmall.Repository
+import java.io.OutputStreamWriter
+import java.nio.file.Path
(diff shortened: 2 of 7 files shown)


build.sbt
```diff
@@ -20,7 +20,7 @@ developers := List(
 resolvers += Resolver.sonatypeRepo("releases")
 libraryDependencies ++= List(
   "org.eclipse.jgit" % "org.eclipse.jgit" % "4.5.4.201711221230-r",
-  "com.geirsson" %% "coursier-small" % "1.0.0-M2",
+  "com.geirsson" %% "coursier-small" % "1.0.0-M4",
```
MasseGuillaume: Add a comment why we need this. GitHub repo link. I guess it's coursier 1.0.0-M4. Why not the latest version?

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -109,7 +120,13 @@ object ScalafixPlugin extends AutoPlugin {
 
   lazy val cli: Either[Throwable, ScalafixInterface] = {
     try {
-      Right(ScalafixInterface.classloadInstance())
+      val dep = new Dependency(
+        "ch.epfl.scala",
+        s"scalafix-cli_${BuildInfo.scala212}",
```
MasseGuillaume: What if we keep the `"g" % "a" %% "v"`syntax and assert that the scalaVersion == BuildInfo.scala212 ? To me this feels more natural, because I always use `%%` for scala library.

olafurpg: This is not the syntax for `scalafixDependencies`, this is the coursier-small API that users won't observe

olafurpg: See `basic/build.sbt` for an example `scalafixDependencies` usage

build.sbt
olafurpg: Done, added a comment.

src/sbt-test/sbt-scalafix/basic/build.sbt
```diff
@@ -5,6 +5,10 @@ inThisBuild(
       "org.scalameta" %% "testkit" % "4.0.0-M6" % Test,
       "org.scalacheck" %% "scalacheck" % "1.13.5" % Test,
       "org.scalatest" %% "scalatest" % "3.0.5" % Test
+    ),
+    scalafixDependencies := List(
+      // Custom rule published to Maven Central https://github.com/olafurpg/example-scalafix-rule
+      "com.geirsson" % "example-scalafix-rule_2.12" % "1.1.1"
```
MasseGuillaume: Should be `"com.geirsson" %% "example-scalafix-rule" % "1.1.1"`

olafurpg: That won't work for the scripted tests on 0.13

MasseGuillaume: Fair point.

PR #11, opened 2018-08-09
Upgrade to Scalafix v0.6.0-M14

olafurpg:
This upgrade introduced a new Java API in a module scalafix-interfaces
that exposes a public API for build tools and IDEs
scalacenter/scalafix#783. This means
sbt-scalafix no longer needs to invoke the fragile
`main(args: Array[String]): Unit` interface. This also means it's
possible to hook into the linter reporting pipeline and use the
sbt logger for diagnostics.


olafurpg:
CI is green meaning tests pass on windows, sbt 0.13 and 1.0 🎉

PR #12, opened 2018-08-09
Throw FeedbackProvidedError to avoid redundant logs

olafurpg:
This makes th output from failed runs cleaner by eliminating redudant
lines. The change is particularly welcome when running aggregated tasks
across multiple projects

Before:
```
[error] TestError
[error] TestError
[error] (a/:sbtfixTest) TestError
[error] (b/:sbtfixTest) TestError
```

After:
```
[error] (a/sbtfixTest) TestError
[error] (b/sbtfixTest) TestError
```

Thanks @jvican for the help finding this.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixFailed.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixFailed.scala
@@ -1,0 +1,8 @@
+package scalafix.sbt
+
+import sbt.FeedbackProvidedException
+import scalafix.interfaces.ScalafixError
+
+final class ScalafixFailed(val errors: List[ScalafixError])
+    extends RuntimeException(errors.mkString(" "))
+    with FeedbackProvidedException
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -336,7 +336,7 @@ object ScalafixPlugin extends AutoPlugin {
 
         val errors = api.runMain(mainArgs)
         if (errors.nonEmpty) {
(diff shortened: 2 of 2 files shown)


src/main/scala/scalafix/sbt/ScalafixFailed.scala
```diff
@@ -0,0 +1,8 @@
+package scalafix.sbt
+
+import sbt.FeedbackProvidedException
```
olafurpg: cc/ @dwijnand 

Why is `sbt.internal.util.FeedbackProvidedException` exposed in the public sbt API via `sbt.FeedbackProvidedException`? Type aliases don't exist in bytecode so
```scala
type FeedbackProvidedException = sbt.internal.util.FeedbackProvidedException
```
will make sbt-scalafix binaries reference `sbt.internal` classes.

Also, why is `FeedbackProvidedException` internal? It's a useful helper for writing custom sbt tasks.

olafurpg: My understanding of the general convention around `internal` packages is that binaries have no guarantees about binary stability. Removing `sbt.internal.util.FeedbackProvidedException` and pointing `sbt.FeedbackProvidedException` somewhere else is a source compatible but binary breaking change.

dwijnand: sbt has done a very poor job in this regard. For instance the entirety of sbt's parser/complete API is "internal" but clearly can't break.

This is just another example of that. TIL `sbt.FeedbackProvidedException`.

olafurpg: Fair enough, too late to change now :)

PR #22, opened 2018-08-27
Add completion for loaded rules (fix scalacenter/scalafix#822)

MasseGuillaume:


src/test/scala/scalafix/internal/sbt/SbtCompletionsSuite.scala
```diff
@@ -32,8 +32,14 @@ class SbtCompletionsSuite extends FunSuite {
   }
   git.tag("v0.1.0")
 
+
+  val exampleDependency = sbt.ModuleID("com.geirsson", "example-scalafix-rule_2.12", "1.2.0")
+  val (_, mainArgs) = scalafix.sbt.ScalafixPlugin.classloadScalafixAPI(ConsoleLogger(), Seq(exampleDependency))
+  val loadedRules = mainArgs.availableRules.asScala.toList
+  loadedRules.foreach(println)
```
MasseGuillaume: This only show built-in rules:

```
ScalafixRule(DisableSyntax)
ScalafixRule(ExplicitResultTypes)
ScalafixRule(NoAutoTupling)
ScalafixRule(NoValInForComprehension)
ScalafixRule(ProcedureSyntax)
ScalafixRule(RemoveUnusedImports)
ScalafixRule(RemoveUnusedTerms)
ScalafixRule(LeakingImplicitClassVal)
ScalafixRule(SyntacticRule)
ScalafixRule(SemanticRule)
```

olafurpg: These two rules here are from example-scalafix-rule:

> ScalafixRule(SyntacticRule)
> ScalafixRule(SemanticRule)

Seems to work as expected.

MasseGuillaume: Oh I see: https://github.com/olafurpg/example-scalafix-rule/blob/master/scalafix/rules/src/main/scala/fix/Examplescalafixrule_v1.scala

It's not picking up v0 however. I guess this is expected as well.

olafurpg: That's expected, only v1 rules can use the service loader framework.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -240,114 +210,57 @@ object ScalafixPlugin extends AutoPlugin {
...
+        case Some(x) =>
+          mainArgs = mainArgs.withConfig(x.toPath)
+        case _ =>
+      }
+      val logger = streams.value.log
+
+      val workingDirectory = baseDirectory.in(ThisBuild).value.toPath
+      val loadedRules = mainArgs.availableRules.asScala.toList
+
+      val inputArgs = new ScalafixCompletions(workingDirectory, loadedRules).parser.parsed
```
MasseGuillaume: @Duhemm @jvican, I would like to have some help here. I get:

```
[error] /home/gui/sbt-scalafix/src/main/scala/scalafix/sbt/ScalafixPlugin.scala:246:47: Illegal dynamic reference: workingDirectory
[error]       val inputArgs = new ScalafixCompletions(workingDirectory, loadedRules).parser.parsed
[error]                                               ^
[error] /home/gui/sbt-scalafix/src/main/scala/scalafix/sbt/ScalafixPlugin.scala:246:65: Illegal dynamic reference: loadedRules
[error]       val inputArgs = new ScalafixCompletions(workingDirectory, loadedRules).parser.parsed
[error]                                                                 ^
[error] two errors found
```

I tried to fix this by nesting `inputTaskDyn` and `taskDyn`. It always end up with an illegal dynamic reference.

MasseGuillaume: This is a dead-end: https://www.scala-sbt.org/1.0/docs/Input-Tasks.html#Constructing+the+Parser
>  Again, we were able to access the current Scala and sbt version for the project because they are settings. Tasks cannot be used to define the parser.

jvican: Yes, I was surprised when you told me you wanted to provide completions for rewrites that are not built-in Scalafix, because that means you need to class load them and all that, which already brings you outside of the settings world.

olafurpg: @MasseGuillaume why do you need to call a task? `scalafixAPI` is a setting https://github.com/scalacenter/sbt-scalafix/blob/7c00166740857df334b90c9a7b8701f66fca3721/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L237-L241

PR #24, opened 2018-09-04
Several UX improvements

olafurpg:
- skip compilation for syntactic rules. This required a significant
  refactoring of how the completion parser is loaded because we cannot
  read setting values from dynamic input tasks.
- use equal sign `=` instead of space ` ` to separate --flag and value
  to avoid ambiguity.
- remove overhead from loading the build, adding sbt-scalafix
  should not slow down `reload`. This means the first `scalafix <TAB>`
  invocation will be slower but I think it's worth it.
- report helpful error message when semanticdb-scalac is disabled
- make completion suggestions print more comfortably on a smaller
  screen. Previously, the suggestions would break my split screen
  making them difficult to read. Now, rule descriptions are printed
  on a separate line making it easier to follow.

Some of the things here are difficult to test automatically. It's very
quick to manually test the plugin however, since the build itself uses
the plugin.

- Fixes https://github.com/scalacenter/scalafix/issues/847
- Fixes https://github.com/scalacenter/scalafix/issues/841
- Fixes https://github.com/scalacenter/scalafix/issues/844


src/main/scala/scalafix/internal/sbt/ScalafixCompletions.scala
```diff
@@ -142,41 +127,43 @@ class ScalafixCompletions(
 
   def hide(p: P): P = p.examples()
 
-  def parser: Parser[Seq[String]] = {
+  def isSemantic(rule: String): Boolean =
```
MasseGuillaume: !

src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
```diff
@@ -0,0 +1,75 @@
...
+        CoursierSmall.fetch(
+          fetchSettings.withDependencies(scalafixCli :: dependencies.toList)
+        )
+      }
+    }
+
+  private val silentCoursierWriter = new OutputStreamWriter(System.out) {
+    override def write(str: String): Unit = {
+      if (str.endsWith(".pom\n") || str.endsWith(".pom.sha1\n")) {
+        () // Ignore noisy "Downloading $URL.pom" logs that appear even for cached artifacts
```
MasseGuillaume: This could swallow useful information. Can you have a regex to select something more precise?

src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
```diff
@@ -0,0 +1,75 @@
...
+      } else {
+        super.write(str)
+      }
+    }
+  }
+
+  private val fetchSettings = new coursiersmall.Settings()
+    .withRepositories(
+      List(
+        Repository.MavenCentral,
```
MasseGuillaume: You should check locally first.

```scala
 List(
  Repository.Ivy2Local,
  Repository.MavenCentral,
  Repository.SonatypeReleases,
  Repository.SonatypeSnapshots
)
```

src/sbt-test/sbt-scalafix/inconfig/test
```diff
@@ -1,5 +1,5 @@
--> ;scalafix --test ;it:scalafix --test
-> it:scalafix
-> it:scalafix --test
--> scalafix --test
+-> ; example/scalafix --test ; example/it:scalafixTest
```
MasseGuillaume: `it:scalafixTest`

src/main/scala/scalafix/internal/sbt/ScalafixCompletions.scala
olafurpg: Removed, it was not used. It was here originally as a dummy implementation before I could upgrade to the latest scalafix.

src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
olafurpg: This filters only the output from coursier, which has limited format so it's unlikely to silence useful information. In case of problems, there will be an exception with rich information about what was not found. This write is only used for interactive feedback.

src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
olafurpg: `MavenCentral` is the first hit, ivy2local is only used with a `publishLocal` version of scalafix-interfaces (which is never in my experience). I just double-checked that it doesn't hit the internet regardless of what order I have (when cached).

src/sbt-test/sbt-scalafix/inconfig/test
olafurpg: Good catch, was a bad sed replace

PR #29, opened 2018-09-21
Skip validation if there are no sources to fix.

olafurpg:
Previously, the semanticdb-scalac and scalacOptions validation ran for
empty root projects with no sources, failing the build when running
scalafix from the root project.  Since there are no sources to fix we
should not run any validation.

Fixes https://github.com/scalacenter/scalafix/issues/884

### DIFF
--- a/src/sbt-test/sbt-scalafix/root-validation/build.sbt
+++ b/src/sbt-test/sbt-scalafix/root-validation/build.sbt
@@ -1,0 +1,20 @@
+inThisBuild(
+  List(
+    scalaVersion := "2.12.6",
+  )
+)
+
+lazy val root = project
+  .in(file("."))
+  .aggregate(lib)
+
+lazy val lib = project
+  .in(file("lib"))
+  .settings(
+    addCompilerPlugin(scalafixSemanticdb),
+    scalacOptions ++= List(
+      "-Ywarn-unused",
+      "-Yrangepos"
+    )
+  )
+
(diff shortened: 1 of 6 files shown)


src/sbt-test/sbt-scalafix/root-validation/build.sbt
```diff
@@ -0,0 +1,20 @@
+inThisBuild(
+  List(
+    scalaVersion := "2.12.6",
+  )
+)
+
+lazy val root = project
```
olafurpg: @abdheshkumar I was able to reproduce the bug with this build.sbt, I got the following error before the fix

```
[info] [error] (Compile / scalafix) scalafix.sbt.InvalidArgument: 2 errors
[info] [error] [E1] The semanticdb-scalac compiler plugin is required to run semantic rules like RemoveUnused.
[info] [error] To fix this problem for this sbt shell session, run `scalafixEnable` and try again.
[info] [error] To fix this problem permanently for your build, add the following settings to build.sbt:
[info] [error]   addCompilerPlugin(scalafixSemanticdb)
[info] [error]   scalacOptions += "-Yrangepos"
[info] [error]
[info] [error] [E2] The Scala compiler option "-Ywarn-unused" is required to use RemoveUnused.
[info] [error] To fix this problem, update your build to use at least one Scala compiler
[info] [error] option that starts with "-Ywarn-unused"
```

After the fix `scalafix` runs successfully.

PR #31, opened 2018-11-20
Added `scalafixResolvers` setting. WIP.

kamilkloch:
See https://github.com/scalacenter/scalafix/issues/866

### DIFF
--- a/src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
+++ b/src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
@@ -1,15 +1,17 @@
 package scalafix.internal.sbt
 
-import com.geirsson.coursiersmall
-import com.geirsson.coursiersmall._
 import java.io.OutputStreamWriter
 import java.net.URLClassLoader
 import java.nio.file.Path
 import java.util.function
 import java.{util => jutil}
+
+import com.geirsson.coursiersmall
+import com.geirsson.coursiersmall._
 import sbt._
-import scala.concurrent.duration.Duration
 import scalafix.sbt.BuildInfo
+
+import scala.concurrent.duration.Duration
 
 object ScalafixCoursier {
   private def scalafixCli: Dependency = new Dependency(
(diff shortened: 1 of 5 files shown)


src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
```diff
@@ -44,24 +45,27 @@ object ScalafixCoursier {
...
+            else ""
+          new Dependency(
+            module.organization,
+            module.name + binarySuffix,
+            module.revision
+          )
+        }
+        CoursierSmall.fetch(
+          fetchSettings
+            .withRepositories(resolvers ++: fetchSettings.repositories)
```
kamilkloch: Note that this breaks a standard sbt behaviour of `scalafixResolvers`: with `scalafixResolvers := Nil` we will still end up with resolver from `fetchSettings.repositories`.

olafurpg: How about renaming the setting to `scalafixCustomResolvers`? I don't see a valid use-case to overwrite the base resolvers.

olafurpg: FYI, I kickstarted a release for coursier-small v1.2.0 https://travis-ci.org/olafurpg/coursier-small/builds/458497906 should be on Maven within the next hr

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -28,6 +28,10 @@ object ScalafixPlugin extends AutoPlugin {
           "For example: scalafix RemoveUnusedImports. " +
           "To run on test sources use test:scalafix."
       )
+    val scalafixCustomResolvers: SettingKey[Seq[Repository]] =
+      settingKey[Seq[Repository]](
+        "Optional list of Repositories used for fetching custom rules."
```
kamilkloch: Any ideas how to improve this doc?

PR #34, opened 2019-01-07
Use custom resolvers

sideeffffect:
fixes https://github.com/scalacenter/scalafix/issues/929

### DIFF
--- a/src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
+++ b/src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
@@ -20,8 +20,8 @@ object ScalafixCoursier {
     BuildInfo.scalafixVersion
   )
 
-  def scalafixCliJars: List[Path] = {
-    CoursierSmall.fetch(fetchSettings.withDependencies(List(scalafixCli)))
+  def scalafixCliJars(repositories: Seq[Repository]): List[Path] = {
+    CoursierSmall.fetch(fetchSettings(repositories, List(scalafixCli)))
   }
   def scalafixToolClasspath(
       deps: Seq[ModuleID],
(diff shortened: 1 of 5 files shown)


sideeffffect:
@olafurpg thanks for merging this! could you please release a new `sbt-scalafix` which contains this fix?
would it be `v0.9.2` or `v0.9.1.1`?

olafurpg:
@sideeffffect there is already a SNAPSHOT release out 0.9.1+2-4639946a-SNAPSHOT. Stable releases happen more infrequently since they require a lot of synchronized work between multiple repos.

sideeffffect:
oh, that's right. I suppose I could use that in the meantime. thanks for reminding me :)

sideeffffect:
> SNAPSHOTs are not documented

https://github.com/scalacenter/sbt-scalafix/pull/35

src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
```diff
@@ -20,8 +20,8 @@ object ScalafixCoursier {
     BuildInfo.scalafixVersion
   )
 
-  def scalafixCliJars: List[Path] = {
-    CoursierSmall.fetch(fetchSettings.withDependencies(List(scalafixCli)))
+  def scalafixCliJars(repositories: Seq[Repository]): List[Path] = {
+    CoursierSmall.fetch(fetchSettings(repositories, List(scalafixCli)))
```
olafurpg: Looking at this change again I'm not convinced this was necessary. The classpath fetched here should only contain scalafix-cli and not custom rules. It's only the "tool classpath" that should use custom resolvers. I had validated that custom resolvers worked as expected so I'm starting to believe something else is going on with your custom nexus repository, does it maybe require authentication?

sideeffffect: no, the Nexus repo in question doesn't require authentication

before this PR was merged, `scalafix` was downloading some (not all, but still some) artifacts from `repo1.maven.org`, even though `scalafixCustomResolvers` was configured to use only the internal Nexus mirror
see the screenshots in the linked issue as proof :wink: 

after this change, `scalafix` abides by the `scalafixCustomResolvers` (for all artifacts, not just some)

maybe my PR changes more things than necessary, I'm not sure, I'm not deeply familiar with the codebase. but some fix was necessary, because `scalafixCustomResolvers` wasn't working correctly before

PR #35, opened 2019-01-08
Add badges

sideeffffect:


olafurpg:
Note that the main documentation for sbt-scalafix is here https://scalacenter.github.io/scalafix/docs/users/installation.html#sbt, not the repo readme.

sideeffffect:
> broken nexus badges

let's wait if shields can fix the bug

> coverage

are you sure we don't want that? I mean, it can't hurt, only help

> users

I actually like this feature. I was able to discover [`cakesolutions/sbt-cake`](https://index.scala-lang.org/cakesolutions/sbt-cake/sbt-cake/1.2.1?target=_2.12_1.0) which I haven't heard of until now

sideeffffect:
@olafurpg are you sure we don't want this in README? I think it would be an improvement :wink:

PR #36, opened 2019-01-17
Upgrade coursier-small to 1.3.3

sideeffffect:
hopefully fixes https://github.com/scalacenter/scalafix/issues/930

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -4,11 +4,8 @@ import java.nio.file.Path
 import java.{util => jutil}
 
 import com.geirsson.coursiersmall.Repository
-import java.io.PrintStream
-
 import sbt.Keys._
-import sbt.internal.sbtscalafix.Compat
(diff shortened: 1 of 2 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -4,11 +4,8 @@ import java.nio.file.Path
 import java.{util => jutil}
 
 import com.geirsson.coursiersmall.Repository
-import java.io.PrintStream
-
 import sbt.Keys._
-import sbt.internal.sbtscalafix.Compat
-import sbt.internal.sbtscalafix.JLineAccess
+import sbt.internal.sbtscalafix.{Compat, JLineAccess}
```
olafurpg: Please try to avoid unrelated diffs like this, it makes it easier to review

olafurpg:
> hopefully fixes scalacenter/scalafix#930

Is there a reason you can't validate locally whether this change fixes the issue?

sideeffffect:
@olafurpg it would probably be possible somehow, but it's always easier to test jars which are already published to Maven central -- especially in a corporate setting, things are always more complicated here

PR #39, opened 2019-02-06
Upgrade to 0.9.4

olafurpg:


olafurpg:
Ignoring flaky windows failure.

bjaglin:
It looks like the `BuildInfo` was not updated like in https://github.com/scalacenter/sbt-scalafix/pull/38/files#diff-a59362d54cf3ed8aeb9c86b2c2eac097 - not sure it's worth an issue since we can't push a patch that wouldn't follow `scalafix` semver anyway.

PR #41, opened 2019-08-22
Bump versions and fix some deprecations

regadas:


src/main/scala/scalafix/sbt/ScalafixEnable.scala
```diff
@@ -57,7 +57,7 @@ object ScalafixEnable {
       )
     } yield setting
 
-    val semanticdbInstalled = extracted.append(settings, s)
+    val semanticdbInstalled = extracted.appendWithoutSession(settings, s)
```
regadas: ahh didn't noticed that this was still being tested against sbt 0.13.x @olafurpg do you want to keep it?

src/main/scala/scalafix/sbt/BuildInfo.scala
```diff
@@ -5,13 +5,11 @@ import java.util.Properties
 object BuildInfo {
 
   def scalafixVersion: String =
-    props.getProperty("scalafixVersion", "0.9.3")
-  def scalafixStableVersion: String =
```
regadas: wasn't used anywhere ¯\_(ツ)_/¯

build.sbt
```diff
@@ -46,7 +46,7 @@ crossScalaVersions := Seq(scala212, scala210)
 sbtVersion in pluginCrossBuild := {
   scalaBinaryVersion.value match {
     case "2.10" => "0.13.17"
-    case "2.12" => "1.2.1"
```
olafurpg: The sbt version here is intentionally old to make sbt-scalafix binary compatible with more sbt versions

src/main/scala/scalafix/sbt/BuildInfo.scala
olafurpg: It may be used by users in the wild, it would be better to deprecate it

src/main/scala/scalafix/sbt/BuildInfo.scala
```diff
@@ -5,13 +5,11 @@ import java.util.Properties
...
-    props.getProperty("scalafixStableVersion", "0.9.3")
+    props.getProperty("scalafixVersion", "0.9.6")
   def scalametaVersion: String =
-    props.getProperty("scalametaVersion", "4.1.0")
+    props.getProperty("scalametaVersion", "4.2.2")
```
olafurpg: These hardcoded version numbers aren't really used since the property file should always be available. We might want to print a warning when the property file is missing saying that we're falling back to an old scalafix version 🤔

src/main/scala/scalafix/sbt/ScalafixEnable.scala
olafurpg: Yes, we should keep this for compatibility with 0.13.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -320,4 +320,6 @@ object ScalafixPlugin extends AutoPlugin {
       }
     }
   }
+
+  final class UninitializedError extends RuntimeException("uninitialized value")
```
olafurpg: What is this used for?

regadas: ah, `UninitalizedError` was deprecated in `2.12.7` so I added it here. It's used here https://github.com/scalacenter/sbt-scalafix/blob/master/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L101

src/main/scala/scalafix/sbt/BuildInfo.scala
regadas: I see, yeah we can do that!

regadas:
@olafurpg ah no worries, no problem at all, is just that we are starting to use scalafix a lot and decided to go through the plugin and saw this minor things.

src/main/scala/scalafix/sbt/BuildInfo.scala
```diff
@@ -25,8 +29,16 @@ object BuildInfo {
...
-        println(s"error: failed to load $path")
+        Logger.error(s"failed to load $path")
     }
     props
   }
+
+  private def property(key: String, default: String): String =
+    Option(props.getProperty(key)).getOrElse {
+      Logger.warn(
+        s"property $key missing; falling back to older version $default"
```
olafurpg:
```suggestion
        s"sbt-scalafix property '$key' missing; falling back to older version '$default'"
```

src/main/scala/scalafix/sbt/ScalafixEnable.scala
```diff
@@ -57,7 +58,7 @@ object ScalafixEnable {
       )
     } yield setting
 
-    val semanticdbInstalled = extracted.append(settings, s)
+    val semanticdbInstalled = Compat.append(extracted, settings, s)
```
olafurpg: :+1:

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
olafurpg: I see, makes sense!

src/main/scala/scalafix/sbt/BuildInfo.scala
regadas: 👍

olafurpg:
Thank you!

PR #42, opened 2019-09-18
Use lazy val instead of def

eatkins:
The projectSettings definition takes about 150ms to run on my laptop
during akka startup. It gets evaluated for each subproject in the build
which adds up. After this change, I found that the average time to load
akka dropped from roughly 19 seconds to 15.5 seconds.


olafurpg:
v0.9.7-1 is on it's way to maven central now

eatkins:
Glad to be of help. I found it by accident :)

PR #89, opened 2020-04-12
become compatible with sbt semanticdbEnabled settings

taisukeoe:
This PullRequest makes SemanticRules work through `semanticdbEnabled` settings which are [introduced in sbt 1.3](https://github.com/sbt/sbt/pull/4410).

With this change, Scalafix users will get the following benefits:

- Being able to add SemanticDB settings in `ThisBuild` scope. (In some cases, it would be handy.)
- Re-using SemanticDB settings of another tools, for sbt-scalafix as well. (It seems to be an expected use case of `semanticdbEnabled` key.)

## Note
While I adds sbt 1.3.x specific scripted test, it seems not to be invoked in sbt 0.13 scenario.
I believe its test is only for sbt 1.3 and can be ignored in sbt 0.13 case. If not, please let me know.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -217,7 +217,7 @@ object ScalafixPlugin extends AutoPlugin {
       shellArgs: ShellArgs,
       config: Configuration
   ): Def.Initialize[Task[Unit]] = Def.taskDyn {
-    val dependencies = libraryDependencies.value
+    val dependencies = allDependencies.value
     val files = filesToFix(shellArgs, config).value
     val withScalaArgs = mainArgs
       .withScalaVersion(scalaVersion.value)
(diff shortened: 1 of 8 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -217,7 +217,7 @@ object ScalafixPlugin extends AutoPlugin {
       shellArgs: ShellArgs,
       config: Configuration
   ): Def.Initialize[Task[Unit]] = Def.taskDyn {
-    val dependencies = libraryDependencies.value
+    val dependencies = allDependencies.value
```
taisukeoe, line 220: Since libraryDependencies will be eventually added to allDependencies, I believe this change will not break sbt-scalafix user codes.

taisukeoe:
Thank you for merging!

olafurpg:
I have kickstarted a v0.9.14 release of scalafix-core and will followup with a sbt-scalafix release once that's done

taisukeoe:
@olafurpg 
Thank you for quick releasing it!
Although I might miss the release, I've sent a PR to update the document as well.
https://github.com/scalacenter/scalafix/pull/1094

PR #90, opened 2020-04-14
Add ThisBuild recommendation settings to error message

taisukeoe:
Following up of #89 , https://github.com/scalacenter/scalafix/pull/1094

Based on [this discussion](https://github.com/scalacenter/scalafix/pull/1094#discussion_r408247753), this PullRequest changes missing SemanticDB error message to ThisBuild configuration example if user sbt version is 1.3 or above.

### DIFF
--- a/src/main/scala/scalafix/sbt/SemanticRuleValidator.scala
+++ b/src/main/scala/scalafix/sbt/SemanticRuleValidator.scala
@@ -1,0 +1,67 @@
+package scalafix.sbt
+
+import java.nio.file.Path
+
+import sbt.librarymanagement.{CrossVersion, ModuleID}
+import scalafix.interfaces.ScalafixArguments
+
+import scala.collection.mutable.ListBuffer
+
+private[sbt] class SemanticRuleValidator(ifNotFound: => SemanticdbNotFound) {
+  def findErrors(
+      files: Seq[Path],
+      dependencies: Seq[ModuleID],
+      args: ScalafixArguments
+  ): Seq[String] = {
+    if (files.isEmpty) Nil
+    else {
+      val errors = ListBuffer.empty[String]
+      val hasSemanticdb =
+        dependencies.exists(_.name.startsWith("semanticdb-scalac"))
+      if (!hasSemanticdb)
+        errors += ifNotFound.message
+      val invalidArguments = args.validate()
+      if (invalidArguments.isPresent)
+        errors += invalidArguments.get().getMessage
+      errors
+    }
+  }
+}
+
+private[sbt] class SemanticdbNotFound(
+    ruleNames: Seq[String],
+    scalaVersion: String,
+    sbtVersion: String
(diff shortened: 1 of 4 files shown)


taisukeoe:
It seems to fail downloading Scalafix cli...
https://travis-ci.org/github/scalacenter/sbt-scalafix/jobs/674979938#L294

src/main/scala/scalafix/sbt/SemanticRuleValidator.scala
```diff
@@ -0,0 +1,67 @@
+package scalafix.sbt
+
+import java.nio.file.Path
+
+import sbt.{CrossVersion, ModuleID}
+import scalafix.interfaces.ScalafixArguments
+
+import scala.collection.mutable.ListBuffer
+
+private[sbt] class SemanticRuleValidator(ifNotFound: => SemanticdbNotFound) {
```
olafurpg, line 10: Can we remove the accessibility restriction and by-name argument?
```suggestion
class SemanticRuleValidator(ifNotFound: SemanticdbNotFound) {
```

taisukeoe: Thank you for pointing out.
By-name argument can be removed since it's not so important (and not good for binary compat?).

IMO `private[sbt]` should be here, since it's not intended to be referenced by users. Otherwise, we may break users code due to changing this internally used class.
What do you think about this point?

olafurpg: We can move this to the internal package if it's internal, my experience is that it's easier for users to see what is internal vs. public when internal code is moved to a separate package instead of adding `private[pkg]` everywhere

olafurpg: https://github.com/scalacenter/sbt-scalafix/tree/master/src/main/scala/scalafix/internal/sbt

taisukeoe: Thank you for confirmation.
While I prefer restriction over convention, I can understand we don't want to mix up two different ways to say *"This is for internal!"*.

I will update this PR shortly.

olafurpg: It's possible to access `private[pkg]` symbols by writing proxy methods in that package. Both approaches are leaky, but using `internal` in the package name at least makes it easier to see in the client implementation or stack traces when the user is accessing internal APIs.

taisukeoe: Thank you for throughout reviewing and merging all my pull requests!

> it easier to see in the client implementation or stack traces when the user is accessing internal APIs.

While I know there are *escape hatches* in both approaches, noticeability in stack traces is a fair point. I'm fine to go the `internal` approach. :)

PR #97, opened 2020-04-26
fix -f/--files

bjaglin:
I might be missing something, but it looks like https://github.com/scalacenter/sbt-scalafix/commit/4dbaf1ece9a67ab686cf5c4e12483298996005de#diff-b6db8e3499369917214568aee2dd0960R222 never worked.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -173,7 +173,6 @@ object ScalafixPlugin extends AutoPlugin {
       config: Configuration
   ): Def.Initialize[Task[Unit]] = Def.task {
     runArgs(
-      shellArgs,
       filesToFix(shellArgs, config).value,
       mainArgs,
       streams.value.log,
@@ -201,7 +200,6 @@ object ScalafixPlugin extends AutoPlugin {
           fullClasspath.value.map(_.data.toPath).asJava
         )
         runArgs(
-          shellArgs,
           files,
           args,
           streams.value.log,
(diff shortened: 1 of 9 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -233,13 +230,8 @@ object ScalafixPlugin extends AutoPlugin {
...
 
-    if (paths.nonEmpty || shellArgs.explicitlyListsFiles) {
-
-      if (paths.lengthCompare(1) > 0) {
-        logger.info(
-          s"Running scalafix on ${paths.size} Scala sources"
-        )
-      }
+    if (paths.nonEmpty) {
+      logger.info(s"Running scalafix on ${paths.size} Scala sources")
```
olafurpg, line 234: This should only print when fixing more than one file
```suggestion
    if (paths.lengthCompare(1) > 0) {
      logger.info(s"Running scalafix on ${paths.size} Scala sources")
    }
```

src/sbt-test/sbt-scalafix/explicit-files/test
```diff
@@ -0,0 +1,13 @@
+> scalafixEnable
```
olafurpg, line 1: This should not be needed since `DisableSyntax` is a syntactic rule. No need to change the test, just FYI

bjaglin:
Thanks for your review! Just pushed an amended commit to address both comments.

bjaglin:
A question about the versioning: since versions of `sbt-scalafix` seem to follow `scalafix-core`, I assume this will get out next time a `scalafix-core` is cut?

olafurpg:
I have wanted to separate the versioning of sbt-scalafix and scalafix for a while so that the two projects can evolve at their own pace. The current release process is quite annoying since I have to wait ~30 minutes to release scalafix to bump up a version in sbt-scalafix.


In the meantime, I pushed a tag for `0.9.15-1` so you can use this fix an unblock further work on making the scalafix task incremental, which I'm really excited about 😄

bjaglin:
Thanks for a cutting a release, really appreciated!

PR #99, opened 2020-05-01
Improve parser UX

bjaglin:
The first time I used scalafix, it was through sbt, and I struggled to get it work on the shell while looking at https://scalacenter.github.io/scalafix/docs/users/installation.html#help.

The biggest frustration was the weird behavior or `rules` (the sbt section does advertize the use of the key/prefix-less syntax, but I missed it as I focused on the `man`-like help rather than the tutorial):
```
[root] $ scalafix --rules A --rules B
[error] (Compile / scalafix) scalafix.sbt.InvalidArgument: Unknown rule '--rules'
```
```
[root] $ scalafix --rules A B
[error] (Compile / scalafix) scalafix.sbt.InvalidArgument: the argument '--rules' requires a value but none was supplied
```

I got even more confused when I saw that tests used that syntax: https://github.com/scalacenter/sbt-scalafix/blob/42a0eb006fc45f2d89b20bd219ad1948bf5a70d2/src/test/scala/scalafix/internal/sbt/SbtCompletionsSuite.scala#L135

So this is an attempt at fixing that and a few other things (see commit messages). Error messages can probably be improved further, but that's how far I got with my current knowledge of the sbt parser lib!

My biggest disappointment is that for key/value extras currently unknown from sbt like `--tool-classpath`, there is currently no way to use space as key/value separator: the parser sends the key and the value as 2 separate arguments to `withParsedArguments` as it cannot tell/know if this is a flag or a key/value arg... Should we just list them all anyway? Is there a way to introspect the metaconfig from `scalafix-cli` to automate that and avoid drifting maybe?

### DIFF
--- a/src/main/scala/scalafix/internal/sbt/ScalafixCompletions.scala
+++ b/src/main/scala/scalafix/internal/sbt/ScalafixCompletions.scala
@@ -15,18 +15,17 @@ class ScalafixCompletions(
 ) {
 
   private type P = Parser[String]
+  private type ArgP = Parser[ShellArgs.Arg]
 
   private val equal: P = token("=").map(_.toString)
   private val string: P = StringBasic
-  private def arg(key: String, value: P): P =
-    (key ~ equal ~ value).map { case a ~ _ ~ c => a + "=" + c }
-  private def arg(key: String, shortKey: String, value: P): P =
-    ((key | shortKey).examples(key) ~ equal ~ value).map {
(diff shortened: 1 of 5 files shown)


olafurpg:
The commit history made it a lot easier to follow the changes, thank you for breaking down the changes 🙏

olafurpg:
Just triggered a 0.9.15-2 release with this change

PR #100, opened 2020-05-03
Add support for rules defined/compiled within the same build

bjaglin:
In our fairly big sbt build, we have decided to collocate custom rule definitions with the sources they run against, using a separate sbt project of the same root build, in order to make the process of adding/updating them as seamless as possible (since they are not meant to be distributed or used anywhere else). It's working, but requires a quite hacky use of `--tool-classpath` which can only be passed once, with a lot of coursier-based code from here inlined.

This is an attempt at bringing the feature upstream (that I refer to in the code as "local rules"), although I assume it's not a very common use-case.

I chose to implement it using an ivy configuration (which might have some side-effects I didn't think of!). That led me to 2 interesting thoughts:
1. That configuration can be used to reference external dependencies as well, as an alternative to `scalafixDependencies` (see 72cd47c). I assume this has a tiny extra cost for sbt builds not running coursier (<1.3 or with >=1.3 with coursier explicitly disabled), since these dependencies end up being fetched/stored twice on disk (once into the ivy cache, once into the coursier one) as far as I understand. But it makes it possible to bring some rules only for some modules (not necessarily `ThisBuild` like `scalafixDependencies`). I am not sure it's worth documenting/advertizing though...
1. At some point, I thought explicit usage of coursier could be removed by relying on sbt-librarymanagement only (helping https://github.com/scalacenter/scalafix/issues/1049 for example), but I realized we need to retrieve dynamic JARs for `dependency:*`, and I wouldn't know how to do that without coursier...

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -151,7 +151,7 @@ object ScalafixPlugin extends AutoPlugin {
         terminalWidth = Some(JLineAccess.terminalWidth)
       ).parser.parsed
 
-      val projectDepsInternal =
+      val projectDepsInternal = products.in(ScalafixConfig).value ++
         internalDependencyClasspath.in(ScalafixConfig).value.map(_.data)
       val projectDepsExternal =
         externalDependencyClasspath
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -21,6 +21,9 @@ object ScalafixPlugin extends AutoPlugin {
   object autoImport {
     val Scalafix = Tags.Tag("scalafix")
 
+    val ScalafixConfig = config("scalafix")
+      .describedAs("Dependencies required for rule execution.")
+
     val scalafix: InputKey[Unit] =
       inputKey[Unit](
         "Run scalafix rule in this project and configuration. " +
@@ -71,8 +74,13 @@ object ScalafixPlugin extends AutoPlugin {
 
   import autoImport._
 
+  override lazy val projectConfigurations: Seq[Configuration] =
+    Seq(ScalafixConfig)
+
   override lazy val projectSettings: Seq[Def.Setting[_]] =
-    Seq(Compile, Test).flatMap(c => inConfig(c)(scalafixConfigSettings(c)))
+    Seq(Compile, Test).flatMap(c => inConfig(c)(scalafixConfigSettings(c))) ++
+      inConfig(ScalafixConfig)(Defaults.configSettings) ++
+      Seq(ivyConfigurations += ScalafixConfig)
 
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     initialize := {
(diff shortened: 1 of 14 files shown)


src/sbt-test/sbt-scalafix/local-rules/build.sbt
```diff
@@ -13,12 +13,13 @@ val rules = project
   .disablePlugins(ScalafixPlugin)
   .settings(
     scalaVersion := Versions.scala212,
+    crossPaths := false,
```
bjaglin, line 16: That does not seem to be enough, as I see `org.scala-lang:scala-library:2.13.1` in `service/Scalafix/externalDependencyClasspath`, which would probably cause all sorts of errors if the rule was not dummy. Investigating.

bjaglin: Actually, I can't write a test to show this is a problem, so I am going to assume it's not. I can find 2 explanations for that (I am not sure about them though):
1. [scala-library is binary-compatible](https://docs.scala-lang.org/overviews/core/binary-compatibility-of-scala-releases.html#concretely)
1. the parent classloader already has the 2.12 library, so the 2.13 will effectively be ignored

bjaglin: Somewhat unrelated to this PR, but looking at how class-loading works for this, I was surprised to see that https://github.com/scalacenter/sbt-scalafix/blob/v0.9.15-2/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala#L22-L27 is not reused across `scalafix` invocations. Would it be possible to cache it via a private setting key maybe? I guess it would speed up scalafix when it's run on many projects in the same build.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -21,6 +21,9 @@ object ScalafixPlugin extends AutoPlugin {
   object autoImport {
     val Scalafix = Tags.Tag("scalafix")
 
+    val ScalafixConfig = config("scalafix")
```
bjaglin, line 24: This is the third type of "scalafix config" after the function returning sbt settings applied to `Compile`/`Test` & the sbt setting referencing the actual configuration file for scalafix... I initially considered naming it `Scalafix` and deleting/renaming the current one 2 lines above since it's not used internally since 4dbaf1ece9a67ab686cf5c4e12483298996005de, but that would be a breaking change.

bjaglin:
I have added a questionable commit (1246ef9) that demonstrates declaration of a custom, local rule within the same project as the target code (honoring the description of the `ScalafixConfig` configuration).

I say questionable, as we don't use this but have a multi-project setup as initially tested, and most importantly this has the strong consequence of coupling the scala version of the project with the one of the rules.

In https://github.com/scalacenter/scalafix/issues/1108, I have noted
> change sbt-scalafix to run scalafix_2.13 by default instead of scalafix_2.12

I guess the "by default" imply that it would be possible to switch version anyway, so maybe it wouldn't be that much of a problem as sbt-scalafix could adapt to the rules scala version.

bjaglin:
Rebased and updated the first commit to clarify and improve the classloading layering.

bjaglin:
Merging, as discussed in https://gitter.im/scalacenter/scalafix?at=5ec3f53d8fe27138ba6e01de

PR #102, opened 2020-05-09
Opt-in caching for scalafix invocations

bjaglin:
See https://gitter.im/scalacenter/scalafix?at=5ea5f83aa634f42f7a0524a8.

It took much more effort than expected, mostly due to the compat issues between sbt versions. I was about to give up and say that caching would be effective in sbt 1.x only, but I think I found a decent way to keep the codebase as common as possible. 

As [classpath-only changes (i.e. no source file updated) do not invalidate the cache](https://github.com/scalacenter/sbt-scalafix/commit/b1bc00380d7f715e1fedac6d088fc6ba6ad9e6b8#diff-8d669a43a554c6e50aa493dff351baefR49-R62) and the implementation is probably rough on the edges anyway (false positives?), I was thinking about leaving this opt-in & advertise it as "experimental" for now.

There will most likely be upcoming Scalafix features that trigger false positive cache hits (any new flag/syntax that references a local/remote resource won't be stamped correctly), which can be a problem if the sbt plugin version evolves independently of the Scalafix version. I guess the right thing to do would be to implement caching in Scalafix itself in the long run.


src/main/scala-sbt-0.13/sbt/internal/sbtscalafix/Caching.scala
```diff
@@ -0,0 +1,42 @@
...
+      format.writes(output.value, t)
+
+    override final type Internal = Array[Byte]
+    override final def convert(keys: Seq[Arg.CacheKey]): Array[Byte] = {
+      val baos = new ByteArrayOutputStream()
+      output.withValue(new JavaOutput(baos)) {
+        keys.foreach(stamp)
+      }
+      Hash.apply(baos.toByteArray)
+    }
```
bjaglin, line 36: this effectively replicates the [1.x behavior](https://github.com/sbt/sbt/commit/b0b9dc5e0f6a13bbba11fbf43b98575e4f73f009) (although with a binary representation and a different hashing algorithm)

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -226,18 +239,78 @@ object ScalafixPlugin extends AutoPlugin {
...
+      }
+      val pathsFilesInfo = FilesInfo
+        .lastModified(paths.map(_.toFile).toSet)
+        .asInstanceOf[FilesInfo[ModifiedFileInfo]]
+
+      // used to signal that one of the argument cannot be reliably stamped
+      object StampingImpossible extends RuntimeException with NoStackTrace
+
+      implicit val stamper = new CacheKeysStamper {
+        override protected def stamp: Arg.CacheKey => Unit = {
```
bjaglin, line 254: The overall stamp is the concatenation of each arg stamp, taken in the order they were set, which depends on the flow of the `scalafix` `InputTask` implementation (so relatively stable). As we control this order (as opposed to the user, which can only affect the order of the `ParsedArgs`), it's not necessary to append the type/name of each arg to disambuiguate 2 args that would have the same value.

src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
```diff
@@ -17,45 +17,54 @@ import scala.util.control.NonFatal
...
 
-  case class Classpath(classpath: Seq[Path]) extends Arg {
+  case class Classpath(classpath: Seq[Path]) extends Arg { //FIXME: with CacheKey {
     override def apply(sa: ScalafixArguments): ScalafixArguments =
       sa.withClasspath(classpath.asJava)
   }
+
+  case object NoCache extends Arg with CacheKey {
+    override def apply(sa: ScalafixArguments): ScalafixArguments =
+      sa // caching is currently implemented in sbt-scalafix itself
```
bjaglin, line 66: I am assuming `--no-cache` could become a `scalafix-cli` argument in the future if/when caching is implemented upstream.

bjaglin:
Rebased against https://github.com/scalacenter/sbt-scalafix/pull/103

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -221,22 +241,127 @@ object ScalafixPlugin extends AutoPlugin {
...
+          case Arg.ToolClasspath(toolClassLoader) =>
+            @tailrec def withParentRec(classLoader: URLClassLoader): Unit = {
+              val urls = classLoader.getURLs
+              // to keep it simple, don't support caching when dealing with directories
+              if (urls.exists(_.toString.endsWith("/")))
+                throw StampingImpossible
```
bjaglin, line 263: once/if https://github.com/scalacenter/sbt-scalafix/pull/100 gets merged, I will implement this, but for now, it's impossible to reach/test as the `toolClasspath` set programmatically by the plugin (i.e. not via a CLI `--toolclasspath`, handled below in `ParsedArgs` as we don't have an explicit parser for it) only contains JARs.

bjaglin:
Rebased against https://github.com/scalacenter/sbt-scalafix/pull/100

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: handled in 8b7cb8f

bjaglin:
Merging, as discussed in https://gitter.im/scalacenter/scalafix?at=5ec3f53d8fe27138ba6e01de

PR #106, opened 2020-05-19
log `scalafix --check` diffs as errors via a sbt logger

bjaglin:
Diffs were dumped on stdout. When scalafix is run in parallel with many other tasks across multiple sub-projects, this makes it much easier to find out which files are the culprits and why, as the logs stand out and are available via `sbt last scalafix`.

Lint errors are already logged properly via https://github.com/scalacenter/sbt-scalafix/blob/v0.9.15-2/src/main/scala/scalafix/internal/sbt/ScalafixLogger.scala (and are not dumped to the `printStream` as far as I can tell, so we don't get duplicates).

### DIFF
--- a/src/main/scala/scalafix/internal/sbt/LoggingOutputStream.scala
+++ b/src/main/scala/scalafix/internal/sbt/LoggingOutputStream.scala
@@ -1,0 +1,43 @@
+package scalafix.internal.sbt
+
+import java.io.{ByteArrayOutputStream, OutputStream}
+
+import sbt.{Level, Logger}
+
+/** Split an OutputStream into messages and feed them to a given logger at a specified level. Not thread-safe. */
+class LoggingOutputStream(
+    logger: Logger,
+    level: Level.Value,
+    separator: String
+) extends OutputStream {
+
+  private val baos = new ByteArrayOutputStream {
+    def maybeStripSuffix(suffix: Array[Byte]): Option[String] = {
+      def endsWithSuffix: Boolean =
+        count >= suffix.length && suffix.zipWithIndex.forall {
+          case (b: Byte, i: Int) =>
+            b == buf(count - separatorBytes.length + i)
+        }
+
+      if (endsWithSuffix)
+        Some(new String(buf, 0, count - separatorBytes.length))
+      else None
+    }
+  }
+
+  private val separatorBytes = separator.getBytes
+  require(separatorBytes.length > 0)
+
+  override def write(b: Int): Unit = {
+    baos.write(b)
+    baos.maybeStripSuffix(separatorBytes).foreach { message =>
+      logger.log(level, message)
+      baos.reset()
+    }
+  }
+}
+
+object LoggingOutputStream {
+  def apply(logger: Logger, level: Level.Value): OutputStream =
+    new LoggingOutputStream(logger, level, System.lineSeparator)
+}
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -1,5 +1,6 @@
 package scalafix.sbt
 
+import java.io.PrintStream
 import java.nio.file.{Path, Paths}
 
 import com.geirsson.coursiersmall.Repository
@@ -168,7 +169,8 @@ object ScalafixPlugin extends AutoPlugin {
         loadedRules = () => scalafixInterface().availableRules(),
         terminalWidth = Some(JLineAccess.terminalWidth)
       ).parser.parsed
-
+      val errorLogger =
+        new PrintStream(LoggingOutputStream(streams.value.log, Level.Error))
       val projectDepsInternal = products.in(ScalafixConfig).value ++
         internalDependencyClasspath.in(ScalafixConfig).value.map(_.data)
       val projectDepsExternal =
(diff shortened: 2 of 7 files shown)


src/main/scala/scalafix/internal/sbt/LoggingOutputStream.scala
```diff
@@ -0,0 +1,43 @@
+package scalafix.internal.sbt
+
+import java.io.{ByteArrayOutputStream, OutputStream}
+
+import sbt.{Level, Logger}
+
+/** Split an OutputStream into messages and feed them to a given logger at a specified level. Not thread-safe. */
+class LoggingOutputStream(
```
bjaglin, line 8: Surprisingly, I couldn't find any helper class in sbt... The best I found was https://github.com/sbt/util/blob/d31b9c509384ecaf37f2ffaec8731250e3c60532/internal/util-logging/src/main/scala/sbt/internal/util/LoggerWriter.scala, which is internal and doesn't match the types we need in that case.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -140,6 +141,8 @@ object ScalafixPlugin extends AutoPlugin {
         loadedRules = () => scalafixArgs().availableRules().asScala,
         terminalWidth = Some(JLineAccess.terminalWidth)
       ).parser.parsed
+      val errorLogger =
+        new PrintStream(LoggingOutputStream(streams.value.log, Level.Error))
```
bjaglin, line 145: I thought about deferring the logging until the point we know the status code, to adjust the level accordingly, but looking at the usage of that `printStream`, I think it's reasonnable to always log as error, except *maybe* for [`--stdout`](https://github.com/scalacenter/scalafix/blob/1564a4b7eec87e16f2c6bce233848c6b542eb966/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala#L262), but that's a lot of custom code for something that I don't see being used through sbt

src/sbt-test/sbt-scalafix/basic/build.sbt
```diff
@@ -25,3 +25,19 @@ lazy val example = project
   )
 
 lazy val tests = project
+
+lazy val checkLogs = taskKey[Unit]("Check that diffs are logged as errors")
+
+checkLogs := {
+  val taskStreams = (example / Compile / scalafix / streams).value
+  val reader = taskStreams.readText(taskStreams.key)
```
bjaglin, line 33: The `lastLogFile` used in a [few scripted](https://github.com/sbt/sbt/blob/4c3cd4e535a89d220a0d20f360f36212ceff9b09/sbt/src/sbt-test/reporter/nowarn/build.sbt#L10-L11) did not work as it seems that the except failure just before this gets called flush out the output... In the end, this looks more readable IMHO.

bjaglin:
The 2.10 build has been failing a few times because of network issues, such as
```
2020-05-19T14:39:40.3511999Z [error] https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/com.dwijnand/sbt-dynver/scala_2.12/sbt_1.0/4.0.0/jars/sbt-dynver.jar: download error: Caught java.io.IOException: Server returned HTTP response code: 504 for URL: https://dl.bintray.com/sbt/sbt-plugin-releases/com.dwijnand/sbt-dynver/scala_2.12/sbt_1.0/4.0.0/jars/sbt-dynver.jar (Server returned HTTP response code: 504 for URL: https://dl.bintray.com/sbt/sbt-plugin-releases/com.dwijnand/sbt-dynver/scala_2.12/sbt_1.0/4.0.0/jars/sbt-dynver.jar) while downloading https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/com.dwijnand/sbt-dynver/scala_2.12/sbt_1.0/4.0.0/jars/sbt-dynver.jar
2020-05-19T14:39:40.5602727Z Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? 
```
I'll retry later.

bjaglin:
> Re the 0.13.x failure. I would be OK with dropping sbt 0.13.x support. Most of the sbt plugin ecosystem has moved on from 0.13 and the users on 0.13 can continue to use older sbt-scalafix versions.

In that case it's just a transient infrastructure issue (bintray? I have seen it today on other scripted tests from other projects), but it did take quite a lot of effort to support 0.13.x in https://github.com/scalacenter/sbt-scalafix/pull/102. Now that it's done, we might as well continue for a while, and revisit the decision next time it's a hassle I guess.

bjaglin:
Rebased against https://github.com/scalacenter/sbt-scalafix/pull/100 & https://github.com/scalacenter/sbt-scalafix/pull/102. Since it was not trivial I added a preliminary commit.

olafurpg:
> In that case it's just a transient infrastructure issue (bintray? I have seen it today on other scripted tests from other projects), but it did take quite a lot of effort to support 0.13.x in #102. Now that it's done, we might as well continue for a while, and revisit the decision next time it's a hassle I guess.

Sounds good 👍

bjaglin:
The infra failure was actually hiding flakiness in the new test case (due to the test protocol, not the implementation), particularly on sbt 0.13.x & windows, but with 13efb64 it's finally green.

bjaglin:
@olafurpg should I go ahead and merge this then? I am wondering if I may, but other PRs also look like easy merges. I am willing to help out maintaining this repo in the foreseeable future.

PR #115, opened 2020-05-28
deprecate auto-exported `Scalafix` tag

bjaglin:
Follow-up of https://github.com/scalacenter/sbt-scalafix/pull/100#discussion_r419377950

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -20,7 +20,7 @@ object ScalafixPlugin extends AutoPlugin {
   override def requires: Plugins = JvmPlugin
 
   object autoImport {
-    @deprecated("not used internally, use your own tag")
+    @deprecated("not used internally, use your own tag", "0.9.17")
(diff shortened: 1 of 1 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -20,6 +20,7 @@ object ScalafixPlugin extends AutoPlugin {
   override def requires: Plugins = JvmPlugin
 
   object autoImport {
+    @deprecated("not used internally, use your own tag")
```
olafurpg, line 23: It's often good to include the "since" version or a date when the deprecation happened, to make it easier to evaluate when to delete the method in the future
```suggestion
    @deprecated("not used internally, use your own tag", "0.9.17")
```

bjaglin: I had it initially it, but I wasn't be able to tell if the next release would be `0.9.16-1` or `0.9.17`. I guess it does not really matter if someone gets a warning talking about `0.9.17` on a potential `0.9.16-n`.

PR #119, opened 2020-05-31
fix ScalafixTestkitPlugin tasks description

bjaglin:


bjaglin:
included in https://github.com/scalacenter/sbt-scalafix/pull/120

PR #120, opened 2020-06-01
Improve/document ScalafixTestkitPlugin usage

bjaglin:


src/sbt-test/sbt-1.3/testkit/build.sbt
```diff
@@ -0,0 +1,38 @@
...
+    scalafixTestkitInputClasspath :=
+      fullClasspath.in(input, Compile).value,
+    scalafixTestkitInputScalacOptions :=
+      scalacOptions.in(input, Compile).value,
+    scalafixTestkitInputScalaVersion :=
+      scalaVersion.in(input, Compile).value,
+    mainClass in (Test, run) := Some("fix.IntputOutputSuite")
+  )
+  .dependsOn(input, rules)
+  .enablePlugins(ScalafixTestkitPlugin)
```
bjaglin, line 38: I am not familiar with Docusaurus, but could we capture that somehow (like https://developer.lightbend.com/docs/paradox/current/directives/snippets.html) into https://github.com/scalacenter/scalafix/blob/master/docs/developers/setup.md#library-api? I guess it's tricky because it's not in the same repo...

src/sbt-test/sbt-scalafix/testkit/tests/src/test/scala/fix/InputOutputSuite.scala
```diff
@@ -0,0 +1,34 @@
+package fix
+
+import scalafix.testkit._
+import scala.util.control.NonFatal
+import org.scalatest.exceptions.TestFailedException
+
+object IntputOutputSuite {
+  def main(args: Array[String]): Unit = {
+    if (Array("--save-expect").sameElements(args)) {
```
olafurpg, line 9: This looks generically useful, would it make sense to expose this somehow as an sbt-scalafix task or command where ScalafixTeskitPlugin is enabled?

src/sbt-test/sbt-1.3/testkit/build.sbt
olafurpg: There's nothing like this out of the box that I'm aware of. It's possible to implement somehing like this on top of mdoc with custom modifiers https://scalameta.org/mdoc/docs/modifiers.html#stringmodifier

src/main/scala/scalafix/sbt/ScalafixTestkitPlugin.scala
```diff
@@ -12,15 +12,25 @@ object ScalafixTestkitPlugin extends AutoPlugin {
   object autoImport {
     val scalafixTestkitInputClasspath =
       taskKey[Classpath]("Classpath of input project")
+    val scalafixTestkitInputScalacOptions =
```
olafurpg, line 15: Nice!

src/sbt-test/sbt-scalafix/testkit/tests/src/test/scala/fix/InputOutputSuite.scala
bjaglin: agreed! https://github.com/scalacenter/scalafix/issues/1147

bjaglin: Actually, it just occured to me that I don't need to inline all this in the scripted, as I should be able to reuse `RuleSuite#main()`.

bjaglin: ... and no! now I remember why I didn't do it - https://github.com/scalacenter/scalafix/blob/17a187ccc3e8d8a6acb7b57b4e8fb6e254e4ecd5/scalafix-tests/unit/src/test/scala/scalafix/tests/rule/RuleSuite.scala#L8-L29 is not part of `testkit`, so it's not available here!

PR #121, opened 2020-06-01
allow usage of rules compiled against 2.11 & 2.13

bjaglin:
Ref https://github.com/scalacenter/scalafix/issues/1108

#### Benefits
1. Rules using the presentation compiler can now work on 2.11 and 2.13 input as long as they are cross-compiled (helping with https://github.com/scalacenter/scalafix/issues/998)
1. Advanced users with [local rules](https://github.com/scalacenter/sbt-scalafix/pull/100) targeting 2.11 or 2.13 code will be able to simplify their build

2.12 remains the default until 1.0 to be backward compatible with external rules (i.e loaded through `scalafixDependencies` or `dependency:`) that are not (yet) cross-compiled.

#### TODO before merging
- [x] ~demonstrate (2) in scripted by removing 2.12/2.13 coexistence hacks:~
  - [x] ~https://github.com/NeQuissimus/sort-imports/pull/66~
  - [x] ~https://github.com/NeQuissimus/sort-imports/pull/67~
  - [x] ~https://github.com/olafurpg/example-scalafix-rule/pull/1~
- [x] ~leverage https://github.com/scalacenter/scalafix/pull/1152 once released~

#### TODO in scalafix repo after merging
- add a section in the developer documentation advertising the benefit of cross-building custom rules (can be run in the same invocation as presentation-compiler-dependant ones? will be needed after version xxx?)
- update https://github.com/scalacenter/scalafix/blob/master/docs/rules/ExplicitResultTypes.md#limited-scala-support once this is merged/released as this might unlock support in 2.11 (but not fully in 2.13 https://github.com/scalacenter/scalafix/issues/998#issuecomment-638675679?)
- document `scalafixBinaryScalaVersion`
- document https://github.com/scalacenter/sbt-scalafix/pull/100 now that support for rules can be written in something else than 2.12 without [hacks](https://github.com/scalacenter/sbt-scalafix/pull/100#discussion_r419169706)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -45,6 +45,12 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional list of custom rules to install from Maven Central. " +
           "This setting is read from the global scope so it only needs to be defined once in the build."
       )
+    val scalafixScalaVersion: SettingKey[String] =
+      settingKey[String](
+        "The full scala version used for the scalafix classpath. Rules must be compiled against that binary version." +
+          s"Defaults to the latest supported 2.12 version (${BuildInfo.scala212}). " +
+          "This setting is read from the global scope so it only needs to be defined once in the build."
```
bjaglin, line 52: We could lift this constraint if it's too restrictive. I added it to avoid messing with the classloader structure right away, as (1) in https://github.com/scalacenter/sbt-scalafix/blob/v0.9.16/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala#L105-L120 could no longer be shared across subprojects if we were to allow per-project scala version for scalafix invocations, which would require some basic caching logic if we don't want performance to regress because of JIT warm-ups.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -119,6 +126,7 @@ object ScalafixPlugin extends AutoPlugin {
     scalafixCaching := false,
     scalafixResolvers := ScalafixCoursier.defaultResolvers,
     scalafixDependencies := Nil,
+    scalafixScalaVersion := BuildInfo.scala212, // will follow scalafixVersion once community rules are cross-built
```
bjaglin, line 129: I wonder if it's not worth making this a `TaskKey` (or `scalafixDependencies` a `SettingKey`?), to be able to support https://github.com/scalacenter/scalafix/issues/998 automatically when there is no external rule with something like:
```suggestion
    scalafixScalaVersion := {
      if (scalafixDependencies.value.isEmpty) scalaVersion.value
      else BuildInfo.scala212
    }
```

This can break invocations with `dependency:`, but maybe that's OK as long as the documentation recommends to force `scalafixScalaVersion := BuildInfo.scala212` until 1.x?

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -45,6 +45,12 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional list of custom rules to install from Maven Central. " +
           "This setting is read from the global scope so it only needs to be defined once in the build."
       )
+    val scalafixScalaVersion: SettingKey[String] =
```
bjaglin, line 48: I initially had a boolean that I could not find a good name for (scalafixScalaVersionMatchesBuild`?), with a default of `false` to keep `2.12` no matter what the project is set up with. I guess it would be simpler as ut really represent the intent, if only we could find a good name for it?

bjaglin:
@olafurpg this is not fully ready but it's taking shape - is that what you had in mind?

bjaglin:
cc @mlachkar since this leverages https://github.com/scalacenter/scalafix/pull/1118

bjaglin:
> I think we also need to have a more complicated rule to cross-compile, that requires code changes (example: with folders _2.11, and 2.12), to give examples to help users cross-compile.

Indeed, advanced rules interfacing with the presentation compiler might need that. I guess the best place for this example would be something along https://github.com/olafurpg/example-scalafix-rule (or a copy of it under `scalacenter` as discussed in https://github.com/olafurpg/example-scalafix-rule/pull/1?), since it's not really related to sbt?

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: > Currently, all rules need to be compiled in 2.12 to be used. Right?

Considering the [git history here](https://github.com/scalacenter/sbt-scalafix/commit/a8937c5398abe791334bf82f6860949b486d1d6f#diff-61ff4bfeba35dc5df102214ffbf71224R16) which shows that sbt-scalafix has always been loading rules (built-in or external) in 2.12 and the repos references in the user documentation (https://github.com/olafurpg/example-scalafix-rule & https://github.com/olafurpg/named-literal-arguments), I think it's fair to assume that, yes.

bjaglin: >  I guess we hope that most rules will be cross-compiled soon.

My suggestion in the comment above was more for the transition period: we can leverage the fact that `scalafix-rules` IS cross-built to 2.11/2.13, so picking it up fixes https://github.com/scalacenter/scalafix/issues/998. However, it's safe to do so if there are no external rules as part of the same invocation. If there are, we can't be sure they are cross-compiled, at least not in the near future (0.9.x).

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
olafurpg: I think it's good to use `scalafixScalaVersion` and we can default it to 2.12 for all project

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
olafurpg: The scala version must match exactly in scalafix-cli_$SCALA_VERSION. I don't think we can use `scalaVersion.value` from the project because that would break if users have a mismatch on the patch version (2.13.1 vs 2.13.2).

There are rare cases where using scalafix 2.13.2 with for example scala version 2.13.3 could cause runtime exceptions. I think we shouldn't worry about them since the cost of building against each exact version is too high when compared to the cost of hitting these rare runtime errors (which can often be worked around on the user side by upgrading to a more recent Scala version).

olafurpg:
Just to be clear, I am a big 👍 for merging this PR unblocking 2.11/2.13 support in sbt-scalafix. I don't think we should block this PR in favor of some more general solution for hypothetical use-cases (example: Metals). The sbt plugin is how 90% of users use Scalafix and I think it's totally fine to have an sbt-only solution. I'm just thinking out loud, feel free to ignore me 😅

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: > I don't think we can use scalaVersion.value from the project because that would break if users have a mismatch on the patch version (2.13.1 vs 2.13.2).

Good point, the scripted doesn't exercise this since we use `scalafix.sbt.BuildInfo.scala*` as `scalaVersion`, I will update that.

Considering we add a `projectVersionToScalafixVersion(String: scalaVersion): String` which takes care of looking up the correct patch version based on the binary version the project is built with, what do you think of my proposal

in 0.9.x
```suggestion
    scalafixScalaVersion := {
      if (scalafixDependencies.value.isEmpty) projectVersionToScalafixVersion(scalaVersion.value)
      else BuildInfo.scala212
    }
```

in 1.x
```suggestion
    scalafixScalaVersion := projectVersionToScalafixVersion(scalaVersion.value)
```

bjaglin:
> Just brainstorming, I am wondering if the user should declare the scalafix scala version in .scalafix.conf instead of build.sbt. This information will eventually be needed by other clients to correctly resolve and fetch external rules.

https://github.com/scalacenter/scalafix/pull/1152 does handle external rules fetching, and looking at the implementation, I am not sure promoting the scalafix version to a configuration key would help. Did you have a use-case in mind?

bjaglin:
Putting this on hold until https://github.com/scalacenter/scalafix/pull/1152 is released, as all the heavy lifting can now be done outside `sbt-scalafix`.

src/sbt-test/skip-windows/caching/build.sbt
```diff
@@ -7,7 +7,7 @@ inThisBuild(
     scalacOptions += "-Yrangepos",
     scalacOptions += "-Ywarn-unused", // for RemoveUnused
     scalafixCaching := true,
-    scalafixDependencies += "com.geirsson" %% "example-scalafix-rule" % "1.2.0"
+    scalafixDependencies += "com.nequissimus" %% "sort-imports" % "0.5.3"
```
bjaglin, line 10: I used another artifact as I wanted to have 2 different revisions published under the same organization/name

src/sbt-test/skip-windows/caching/test
```diff
@@ -302,7 +302,7 @@ $ copy-file files/Valid.scala src/main/scala/Valid.scala
 > set scalafixDependencies in ThisBuild := Nil
 > scalafix --check ProcedureSyntax
 > reload plugins
-> 'set dependencyOverrides += "ch.epfl.scala" % "scalafix-interfaces" % "0.9.13"' // anything else than current
+> 'set dependencyOverrides += "ch.epfl.scala" % "scalafix-interfaces" % "0.9.17+7-a1b6d972-SNAPSHOT"' // anything else than current
```
bjaglin, line 305: prevents `java.lang.NoSuchMethodError: scalafix.interfaces.Scalafix.fetchAndClassloadInstance(Ljava/lang/String;Ljava/util/List;)Lscalafix/interfaces/Scalafix;` - it's expected that the user can't downgrade the scalafix version

bjaglin:
Now that cross-publishing of some external rules (thanks @mlachkar) & https://github.com/scalacenter/scalafix/pull/1152 are available, I have reworked the PR and it is now ready for a proper review!

* I have split the PR into several commits to make it easier to see what's going on, but the biggest change is usage of https://github.com/scalacenter/scalafix/pull/1152 - make sure to check the commit message to follow what's going on with class loaders.
* I decided not to pursue https://github.com/scalacenter/sbt-scalafix/pull/121#discussion_r435152906 as I am worried it's too magic, and can break usage of `dependency:` rules. I will explicitly mention in the docs of `ExplicitResultType` the actions to make it available on 2.11 & 2.13 and the limitations of matching the `scalacVersion`.
* I have renamed `scalafixScalaVersion` to `scalafixScalaBinaryVersion` as a follow-up to https://github.com/scalacenter/sbt-scalafix/pull/121#discussion_r435117738.

bjaglin:
> Just brainstorming, I am wondering if the user should declare the scalafix scala version in .scalafix.conf instead of build.sbt thinking This information will eventually be needed by other clients to correctly resolve and fetch external rules. I think it would be nice if we can keep the build clients thin (sbt-scalafix, gradle-scalafix, metals-scalafix,...) and move as much heavy lifting as possible behind scalafix-interfaces.

I agree that it's a bit strange to have it in `build.sbt` (but it's probably the same if it were in `.scalafix.conf`) as the 2.11/2.13 developer will face a dilemna: use ExplicitResultTypes or external rules not cross-published yet?

scalafix-interfaces could abstract the scala version(s) it uses completely with a best-effort strategy, by trying to match the provided scala version for rules that are available for that binary version, and to stick to 2.12 for the others (effectively doing 2 separate runs). That does not sound particularly easy to do though.

build.sbt
```diff
@@ -52,8 +52,8 @@ scalaVersion := scala212
 crossScalaVersions := Seq(scala212, scala210)
 sbtVersion in pluginCrossBuild := {
   scalaBinaryVersion.value match {
-    case "2.10" => "0.13.17"
-    case "2.12" => "1.2.1"
+    case "2.10" => "0.13.18"
+    case "2.12" => "1.2.8"
```
olafurpg, line 56: IIRC, you usually want to target the lowest possible version here to avoid binary compatibility issues on older sbt versions. I’ve had issues in the past because sbt 1.x releases are always backwards compatible but not forwards compatible

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -67,6 +68,12 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional list of custom rules to install from Maven Central. " +
           "This setting is read from the global scope so it only needs to be defined once in the build."
       )
+    val scalafixScalaBinaryVersion: SettingKey[String] =
+      settingKey[String](
+        "The Scala binary version used for scalafix execution. Defaults to 2.12. " +
+          s"Rules must be compiled against that binary version. Advanced semantic rules such as " +
+          s"ExplicitResultTypes require this to match the Scala version used for compilation."
```
olafurpg, line 75: The last sentence may be a bit misleading since it indicates this version must match the scala version exactly (MAJOR.MINOR.PATCH), not just the binary version (MAJOR.MINOR).

src/sbt-test/skip-windows/caching/build.sbt
```diff
@@ -7,7 +7,8 @@ inThisBuild(
     scalacOptions += "-Yrangepos",
     scalacOptions += "-Ywarn-unused", // for RemoveUnused
     scalafixCaching := true,
-    scalafixDependencies += "com.geirsson" %% "example-scalafix-rule" % "1.2.0"
+    scalafixDependencies += "com.nequissimus" %% "sort-imports" % "0.5.3",
```
olafurpg, line 10: Is there a reason you can use the example scala rule? Typically you want to avoid depending on 3rdparty dependencies like this in your tests so that you have more control over everything.

src/main/scala/scalafix/internal/sbt/ScalafixInterfacesClassloader.scala
```diff
@@ -1,19 +0,0 @@
-package scalafix.internal.sbt
-
-/**
-  * A classloader that shares only scalafix-interfaces classes from the parent classloader.
-  *
-  * This classloader is intended to be used as a parent when class-loading scalafix-cli.
-  * By using this classloader as a parent, it's possible to cast runtime instances from
-  * the scalafix-cli classloader into `scalafix.interfaces.Scalafix` from this classlaoder.
-  */
-class ScalafixInterfacesClassloader(parent: ClassLoader)
```
olafurpg, line 10: Nice! Good riddance

src/sbt-test/skip-windows/caching/build.sbt
bjaglin: I mentioned the reason in [the commit message](https://github.com/scalacenter/sbt-scalafix/pull/121/commits/d6124f0445339d8ca12c968263e968b15d204e68). I wanted to move from `com.geirsson` to `ch.epfl.scala` everywhere, but couldn't do it here as I needed 2 revisions (see usage of `set scalafixDependencies` in the corresponding `test`). But I agree it does mot make sence to move from one third party (even if it's yours) to another. I'll revert changes here in that specifc reference and in `test`.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: Indeed, I am rephrasing to
```
        "The Scala binary version used for scalafix execution. Defaults to 2.12. " +
          s"Rules must be compiled against that binary version, or for advanced rules such as " +
          s"ExplicitResultTypes which have a full cross-version, against the corresponding full" +
          s"version that scalafix is built against."
```

build.sbt
bjaglin: I also ran into https://github.com/sbt/sbt/issues/5049 on another project, that's why I didn't bump the minor, just the patch release. The problem is that these versions are not required for building, but as mentioned in 6e70250, for [scripted](https://github.com/sbt/sbt/blob/924d8e75cfb0f77da706925e5eebbe85cda7d868/main/src/main/scala/sbt/ScriptedPlugin.scala#L61) to build 2.13 (I tried to use 2.11 there but ran into another issue).

AFAIK only the 0.13 bump was necessary, so I'll try to keep 1.2.1 to limit the risks.

bjaglin:
Amended the commits based on the feedback, overall diff: https://github.com/scalacenter/sbt-scalafix/compare/5316d91..fb96378

src/main/scala/scalafix/internal/sbt/ScalafixInterfacesClassloader.scala
bjaglin: Well, it just moved https://github.com/scalacenter/scalafix/blob/v0.9.17/scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixInterfacesClassloader.java :)

build.sbt
bjaglin: > AFAIK only the 0.13.18 bump was necessary, so I'll keep 1.2.1 to limit the risks.

Actually, 1.2.7 is required to build 1.13 sources (as it's the first version that brings a Zinc with a compiler-bridge published for 2.13).

bjaglin: Nevermind, I figured out I could just override `scriptedSbt` to leave `sbtVersion` untouched!

src/main/scala/scalafix/internal/sbt/ScalafixInterfacesClassloader.scala
olafurpg: Right! It's still nice to have it out of sbt-scalafix so this class doesn't need to be copy-pasted for every client 😊

PR #123, opened 2020-06-07
deprecate usage of coursiersmall type in scalafixResolvers

bjaglin:
This is breaking binary compatibility, but preserves source compatibility via (deprecated) implicit conversion, which is always in scope for `*.sbt` sources, but require an explicit import for `*.scala`
sources, such as `import scalafix.sbt.ScalafixPlugin.autoImport._`.

That means that most 99% of direct sbt-scalafix users should be fine, but any plugin published against sbt-scalafix will need to be rebuilt.

Usage of coursierapi is a bit more cumbersome for sbt 0.13.x because the Java8 static methods in Repository cannot be used from Scala 2.10... As the scripted shows, it's still possible to setup resolvers without these helpers.

### DIFF
--- a/src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
+++ b/src/main/scala/scalafix/internal/sbt/ScalafixCoursier.scala
@@ -5,7 +5,6 @@ import java.util.function
 import java.{util => jutil}
 
 import coursierapi._
-import com.geirsson.{coursiersmall => cs}
 import sbt._
 import scalafix.sbt.BuildInfo
 import scala.collection.JavaConverters._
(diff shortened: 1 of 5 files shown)


bjaglin:
It didn't even cross my mind as I was deep into coursier-interface, but maybe `sbt.librarymanagement.Repository` is actually the type we want (since `sbt.librarymanagement.ModuleID` is used in `scalafixDependencies`)?

bjaglin:
> I created coursier-small before coursier-interface existed.

I figured that out by looking at the git history (which by the wat is very rich, it's really nice for newcomers). Going back to my previous question: do you remember why you chose to expose types from there instead of mapping `sbt.librarymanagement.*` ?

olafurpg:
There was no good reason to do that. I'm not familiar with `sbt.librarymanagement` and I assumed for some reason that the APIs had changed between 0.13 and 1.x

bjaglin:
Thanks! I'll check if it's possible to use it and amend this in that case (to keep coursier an implementation detail).

PR #125, opened 2020-06-09
support sbt 1.3.x SemanticdbPlugin in ScalafixTestkitPlugin (take 2)

bjaglin:
The test for 4278a1f was too weak as it was not triggering the (lazy) lookup of semanticdb files.

This replicates [the behavior of `scalafix-cli`](https://github.com/scalacenter/scalafix/blob/v0.9.17/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala#L316-L341), inspecting `scalacOptions` to detect a potential custom classpath for semanticdb files (set by default as `semanticdbIncludeInJar := false`).

https://github.com/sbt/sbt/blob/v1.3.12/main/src/main/scala/sbt/Defaults.scala#L469

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixTestkitPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixTestkitPlugin.scala
@@ -33,11 +33,16 @@ object ScalafixTestkitPlugin extends AutoPlugin {
       scalafixTestkitInputScalaVersion := scalaVersion.value,
       resourceGenerators.in(Test) += Def.task {
         val props = new java.util.Properties()
+        val targetrootClasspath = semanticdbOption(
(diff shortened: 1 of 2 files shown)


bjaglin:
Even though https://github.com/scalacenter/scalafix/blob/6811967f6e6059a59d8b04e7f2b76eb3c3ddcfbc/scalafix-testkit/src/main/scala/scalafix/testkit/TestkitProperties.scala#L15-L17 mentions a "single directory", it's quite clear about the requirement, so I think this is better to have this in sbt-scalafix?

bjaglin:
@olafurpg not sure what happened with GitHub, but you appear twice as reviewer after I requested it... so just in case, I am pinging directly if you want to have a look

PR #126, opened 2020-06-13
Introduce scalafixAll InputTask

bjaglin:
Docs companion: https://github.com/scalacenter/scalafix/pull/1160

Similarly to [`scalafmtAll`](https://github.com/scalameta/sbt-scalafmt/blob/v2.4.0/plugin/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala#L436) (except that the implementation here is more tricky as we deal with an `InputTask` for which [`.all`](https://www.scala-sbt.org/1.x/docs/Tasks.html#Getting+values+from+multiple+scopes) does not work), this allows a user to easily and efficiently run Scalafix across configurations for a given project (or of course for all projects/configurations with root project aggregation)

I saw there [were previous discussions/opinions about that](https://github.com/scalacenter/scalafix/pull/587#issuecomment-365218154) (I should have opened an issue to discuss that beforehand). From my side I see no reason not to do it, especially because implementation is hard(er) downstream but simple here (or at least readable & compact), and because it allows `scalafix --check` to run in parallel over all configurations (rather than in sequential as documented in https://scalacenter.github.io/scalafix/docs/users/installation.html#enforce-in-ci), so it's not just sugar.

There has already been work downstream to bring that to users through a command (like https://github.com/alejandrohdezma/sbt-fix#running-scalafix-in-all-configurations), but refactoring the code here allows us to expose an `InputTask`, with the same auto-completion as the existing `scalafix`, which can't be done downstream at the moment.

The first commit is preliminary, separating parsing & core logic from `scalafix` so that the output of the parser is reused across configurations during a `scalafixAll` invocation, and getting rid of a workaround.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -7,9 +7,10 @@ import com.geirsson.coursiersmall.Repository
 import sbt.KeyRanks.Invisible
 import sbt.Keys._
 import sbt._
+import sbt.complete._
+import sbt.internal.sbtscalafix.Caching._
 import sbt.internal.sbtscalafix.{Compat, JLineAccess}
 import sbt.plugins.JvmPlugin
-import sbt.internal.sbtscalafix.Caching._
 import scalafix.interfaces.ScalafixError
 import scalafix.internal.sbt._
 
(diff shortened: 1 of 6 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -124,12 +120,28 @@ object ScalafixPlugin extends AutoPlugin {
   )
 
   lazy val stdoutLogger = Compat.ConsoleLogger(System.out)
-  private var workingDirectory = file("").getAbsoluteFile.toPath
-  private var scalafixInterface: () => ScalafixInterface =
-    () => throw new UninitializedError
+
+  private val scalafixInterface: Def.Initialize[() => ScalafixInterface] =
```
bjaglin, line 124: this is effectively like a `Setting`, so the function is created at sbt load time, but evaluated at task execution time, and since it's backed by a `LazyValue`, it's memoized.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -158,20 +170,49 @@ object ScalafixPlugin extends AutoPlugin {
 
   }
 
+  private def scalafixAllInputTask(): Def.Initialize[InputTask[Unit]] =
+    // workaround https://github.com/sbt/sbt/issues/3572 by invoking directly what Def.inputTaskDyn would via macro
```
olafurpg, line 174: Awesome to see a workaround for this issue, nice job tracking down the issue!

src/sbt-test/skip-windows/caching/test
```diff
@@ -205,6 +205,24 @@ $ exec chmod 000 src/test/scala/Valid.scala
 $ delete src/main/scala
 $ delete src/test/scala
 
+# scalafixAll and scalafix should share the same cache per configuration
```
olafurpg, line 208: 👏

PR #127, opened 2020-06-13
make local dev possible with JDK11

bjaglin:
Follow-up of https://github.com/scalacenter/sbt-scalafix/commit/5f3934a1b300db5e3490154acc5aec9554a3386d which was not enforced in CI and where I missed the scripted sources.

### DIFF
--- a/src/sbt-test/sbt-scalafix/cross-build/build.sbt
+++ b/src/sbt-test/sbt-scalafix/cross-build/build.sbt
@@ -22,7 +22,7 @@ lazy val root = project
   )
 
 lazy val scala210 = project.settings(
-  scalaVersion := "2.10.5"
+  scalaVersion := "2.10.7"
 )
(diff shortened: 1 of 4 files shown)


src/sbt-test/sbt-scalafix/cross-build/build.sbt
```diff
@@ -22,7 +22,7 @@ lazy val root = project
   )
 
 lazy val scala210 = project.settings(
-  scalaVersion := "2.10.5"
+  scalaVersion := "2.10.7"
```
bjaglin, line 25: prevents `error while loading package, Missing dependency 'object java.lang.Object in compiler mirror', required by /home/runner/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.10.5.jar(scala/package.class)` on CI

PR #128, opened 2020-06-16
test/document usage of local 2.13 rules

bjaglin:
Somehow, [this got lost](https://github.com/scalacenter/sbt-scalafix/compare/5316d91..fd97d1a#diff-989d991e5bc1817b8d9f0b2eef7af260) when I [addressed @olafurpg's comments](https://github.com/scalacenter/sbt-scalafix/pull/121#issuecomment-643939807). It demonstrates the second goal of https://github.com/scalacenter/sbt-scalafix/pull/121

> 2. Advanced users with local rules targeting 2.11 or 2.13 code will be able to simplify their build

### DIFF
--- a/src/sbt-test/sbt-scalafix/local-rules/build.sbt
+++ b/src/sbt-test/sbt-scalafix/local-rules/build.sbt
@@ -10,15 +10,17 @@ inThisBuild(
     // error: Static methods in interface require -target:jvm-1.8
     scalafixResolvers := List(
       coursierapi.MavenRepository.of("https://repo1.maven.org/maven2")
-    )
+    ),
+    scalaVersion := "2.13.0", // out of sync with scalafix.sbt.BuildInfo.scala213 on purpose
+    scalafixScalaBinaryVersion :=
+      // this should be the default in sbt-scalafix 1.0
+      CrossVersion.binaryScalaVersion(scalaVersion.value)
   )
 )
 
 val rules = project
   .disablePlugins(ScalafixPlugin)
   .settings(
-    scalaVersion := Versions.scala212,
-    crossPaths := false,
     libraryDependencies += "ch.epfl.scala" %% "scalafix-core" % Versions.scalafixVersion,
     libraryDependencies += "joda-time" % "joda-time" % "2.10.6"
   )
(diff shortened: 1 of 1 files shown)


src/sbt-test/sbt-scalafix/local-rules/build.sbt
```diff
@@ -10,28 +10,28 @@ inThisBuild(
...
     libraryDependencies += "ch.epfl.scala" %% "scalafix-core" % Versions.scalafixVersion,
     libraryDependencies += "joda-time" % "joda-time" % "2.10.6"
   )
 
 val service = project
   .dependsOn(rules % ScalafixConfig)
   .settings(
-    scalaVersion := Versions.scala213,
-    libraryDependencies += "com.nequissimus" % "sort-imports_2.12" % "0.5.0" % ScalafixConfig
+    libraryDependencies += "com.nequissimus" % "sort-imports" % "0.5.2" % ScalafixConfig
```
bjaglin, line 31: https://github.com/NeQuissimus/sort-imports/pull/67

src/sbt-test/sbt-scalafix/local-rules/build.sbt
```diff
@@ -10,28 +10,28 @@ inThisBuild(
...
     libraryDependencies += "ch.epfl.scala" %% "scalafix-core" % Versions.scalafixVersion,
     libraryDependencies += "joda-time" % "joda-time" % "2.10.6"
   )
 
 val service = project
   .dependsOn(rules % ScalafixConfig)
   .settings(
-    scalaVersion := Versions.scala213,
-    libraryDependencies += "com.nequissimus" % "sort-imports_2.12" % "0.5.0" % ScalafixConfig
+    libraryDependencies += "com.nequissimus" %% "sort-imports" % "0.5.2" % ScalafixConfig
```
bjaglin, line 31: https://github.com/NeQuissimus/sort-imports/pull/67

bjaglin: By the way, I am using an external rule here as I need multiple rules, so I can't stick to `example-scalafix-rule`.

jrudolph:
I ran into this as well following the instructions from https://scalacenter.github.io/scalafix/docs/developers/local-rules.html. Might make sense to update the instructions there as well.

bjaglin:
@jrudolph do you mean that https://scalacenter.github.io/scalafix/docs/developers/local-rules.html#prerequisite should also suggest setting the Scala version at the build level?

jrudolph:
> @jrudolph do you mean that [scalacenter.github.io/scalafix/docs/developers/local-rules.html#prerequisite](https://scalacenter.github.io/scalafix/docs/developers/local-rules.html#prerequisite) should also suggest setting the Scala version at the build level?

Doh, I missed that part at the top :) Maybe it's enough to include the error message that you will get if you don't do it there which was a generic "Provider ... could not be instantiated". Maybe a safeguard could be built into the scalafix invocation to avoid that the classpath contains incompatible classes/jars (e.g. just look for wrong "scala_2.13" in classpath or jar file names)? Maybe that would probably not only be useful for the local rules scenario as soon as cross publishing will be more common.

PR #139, opened 2020-07-04
caching: don't forget previous invocations

bjaglin:
This allows cache accumulation after successive runs of scalafix with different rules, which is common as it's currently the only
way to safely run several semantic rules in order.

Note that the cache remains incremental only for files, so if a file is stale for a single rule, all rules will be executed on this file.


bjaglin:
This is particularly useful when combined with [`scalafixOnCompile := true`](https://github.com/scalacenter/sbt-scalafix/pull/140). Without this, a single `scalafixAll RandomRuleIAmTrying` would invalidate all caches.

bjaglin:
Rebased against https://github.com/scalacenter/sbt-scalafix/pull/145 - I thought it would make implementation easier, but it does not.

bjaglin:
@olafurpg I imagine the implementation is getting quite complex to review, but maybe you want to have a look at the tests at least? Thanks!

olafurpg (approved):
LGTM 👍

PR #140, opened 2020-07-05
scalafixOnCompile setting key

bjaglin:
Users can control whether Scalafix should be run (with the rules defined in `.scalafix.conf`) as part of `compile,`at the project & configuration level.

The value of `scalafixOnCompile` is ignored when invoking directly `scalafix` or `scalafixAll`, as the CLI arguments (if there are any) take precedence over the default rules. That means for example that `scalafix --check` is safe to run even though `scalafixOnCompile := true`, as there will not be any rewrite triggered by the implicit compilation.

Enabling `scalafixOnCompile` wil automatically turn on caching, which can be disabled by an explicit `scalafixCaching := false`.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -32,7 +32,7 @@ object ScalafixPlugin extends AutoPlugin {
     val scalafix: InputKey[Unit] =
       inputKey[Unit](
         "Run scalafix rule(s) in this project and configuration. " +
-          "For example: scalafix RemoveUnusedImports. " +
+          "For example: scalafix RemoveUnused. " +
           "To run on test sources use test:scalafix or scalafixAll. " +
           "When invoked directly, prior compilation will be triggered for semantic rules."
       )
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -46,12 +46,13 @@ object ScalafixPlugin extends AutoPlugin {
     val scalafixOnCompile: SettingKey[Boolean] =
       settingKey[Boolean](
         "Run Scalafix rule(s) declared in .scalafix.conf on compilation and fail on lint errors. " +
-          "Off by default."
+          "Off by default. When enabled, caching will be automatically activated, " +
+          "but can be disabled with `scalafixCaching := false`."
       )
 
     val scalafixCaching: SettingKey[Boolean] =
       settingKey[Boolean](
-        "Cache scalafix invocations (off by default, still experimental)."
+        "Cache scalafix invocations (off by default, on if scalafixOnCompile := true)."
       )
 
     import scala.language.implicitConversions
@@ -179,7 +180,6 @@ object ScalafixPlugin extends AutoPlugin {
 
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     scalafixConfig := None, // let scalafix-cli try to infer $CWD/.scalafix.conf
-    scalafixCaching := false,
     scalafixOnCompile := false,
     scalafixResolvers := Seq(
       Repository.ivy2Local(),
@@ -354,10 +354,9 @@ object ScalafixPlugin extends AutoPlugin {
           scalafixResolvers.in(ThisBuild).value,
           projectDepsInternal
         )
+        val cachingRequested = scalafixCaching.or(scalafixOnCompile).value
         val maybeNoCache =
-          if (shell.noCache || !scalafixCaching.in(config).value)
-            Seq(Arg.NoCache)
-          else Nil
+          if (shell.noCache || !cachingRequested) Seq(Arg.NoCache) else Nil
         val mainInterface = mainInterface0
           .withArgs(maybeNoCache: _*)
           .withArgs(
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -33,13 +33,22 @@ object ScalafixPlugin extends AutoPlugin {
       inputKey[Unit](
         "Run scalafix rule(s) in this project and configuration. " +
           "For example: scalafix RemoveUnusedImports. " +
-          "To run on test sources use test:scalafix or scalafixAll."
+          "To run on test sources use test:scalafix or scalafixAll. " +
+          "When invoked directly, prior compilation will be triggered for semantic rules."
       )
     val scalafixAll: InputKey[Unit] =
       inputKey[Unit](
         "Run scalafix rule(s) in this project, for all configurations where scalafix is enabled. " +
-          "Compile and Test are enabled by default, other configurations can be enabled via scalafixConfigSettings."
+          "Compile and Test are enabled by default, other configurations can be enabled via scalafixConfigSettings. " +
+          "When invoked directly, prior compilation will be triggered for semantic rules."
       )
+
+    val scalafixOnCompile: SettingKey[Boolean] =
+      settingKey[Boolean](
+        "Run Scalafix rule(s) declared in .scalafix.conf on compilation and fail on lint errors. " +
+          "Off by default."
+      )
+
     val scalafixCaching: SettingKey[Boolean] =
       settingKey[Boolean](
         "Cache scalafix invocations (off by default, still experimental)."
@@ -87,6 +96,17 @@ object ScalafixPlugin extends AutoPlugin {
     def scalafixConfigSettings(config: Configuration): Seq[Def.Setting[_]] =
       Seq(
         scalafix := scalafixInputTask(config).evaluated,
+        compile := Def.taskDyn {
+          val oldCompile =
+            compile.value // evaluated first, before the potential scalafix evaluation
+          val runScalafixAfterCompile =
+            scalafixOnCompile.value && !scalafixRunExplicitly.value
+          if (runScalafixAfterCompile)
+            scalafix
+              .toTask("")
+              .map(_ => oldCompile)
+          else Def.task(oldCompile)
+        }.value,
         // In some cases (I haven't been able to understand when/why, but this also happens for bgRunMain while
         // fgRunMain is fine), there is no specific streams attached to InputTasks, so  we they end up sharing the
         // global streams, causing issues for cache storage. This does not happen for Tasks, so we define a dummy one
@@ -160,6 +180,7 @@ object ScalafixPlugin extends AutoPlugin {
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     scalafixConfig := None, // let scalafix-cli try to infer $CWD/.scalafix.conf
     scalafixCaching := false,
+    scalafixOnCompile := false,
     scalafixResolvers := Seq(
       Repository.ivy2Local(),
       Repository.central(),
@@ -394,16 +415,22 @@ object ScalafixPlugin extends AutoPlugin {
         new SemanticdbNotFound(ruleNames, scalaVersion.value, sbtVersion.value)
       ).findErrors(files, dependencies, withScalaInterface)
       if (errors.isEmpty) {
-        Def.task {
+        val task = Def.task {
+          // passively consume compilation output without triggering compile as it can result in a cyclic dependency
+          val classpath =
+            dependencyClasspath.in(config).value.map(_.data.toPath) :+
+              classDirectory.in(config).value.toPath
           val semanticInterface = withScalaInterface.withArgs(
             Arg.Paths(files),
-            Arg.Classpath(fullClasspath.in(config).value.map(_.data.toPath))
+            Arg.Classpath(classpath)
           )
           runArgs(
             semanticInterface,
             streams.in(config, scalafix).value
           )
         }
+        if (scalafixRunExplicitly.value) task.dependsOn(compile.in(config))
+        else task
       } else {
         Def.task {
           if (errors.length == 1) {
@@ -536,6 +563,16 @@ object ScalafixPlugin extends AutoPlugin {
       () // do nothing
     }
   }
+
+  // Controls whether scalafix should depend on compile (true) & whether compile may depend on
+  // scalafix (false), to avoid cyclic dependencies causing deadlocks during executions (as
+  // dependencies come from dynamic tasks).
+  private val scalafixRunExplicitly: Def.Initialize[Task[Boolean]] =
+    Def.task {
+      executionRoots.value.exists { root =>
+        Seq(scalafix.key, scalafixAll.key).contains(root.key)
+      }
+    }
 
   private def isScalaFile(file: File): Boolean = {
     val path = file.getPath
--- a/src/sbt-test/sbt-scalafix/scalafixOnCompile/.scalafix.conf
+++ b/src/sbt-test/sbt-scalafix/scalafixOnCompile/.scalafix.conf
@@ -1,0 +1,2 @@
+rules = [DisableSyntax, RemoveUnused]
+DisableSyntax.noNulls = true
--- a/src/sbt-test/sbt-scalafix/scalafixOnCompile/build.sbt
+++ b/src/sbt-test/sbt-scalafix/scalafixOnCompile/build.sbt
@@ -1,0 +1,23 @@
+import _root_.scalafix.sbt.{BuildInfo => Versions}
+
+inThisBuild(
+  Seq(
+    scalaVersion := Versions.scala212,
+    scalacOptions ++= List(
+      "-Yrangepos",
+      "-Ywarn-unused-import"
+    )
+  )
+)
+lazy val lint = project
+  .settings(
+    addCompilerPlugin(scalafixSemanticdb)
+  )
+
+lazy val rewrite = project
+  .configs(IntegrationTest)
+  .settings(
+    Defaults.itSettings,
+    inConfig(IntegrationTest)(scalafixConfigSettings(IntegrationTest)),
+    addCompilerPlugin(scalafixSemanticdb)
+  )
--- a/src/sbt-test/sbt-scalafix/scalafixOnCompile/lint/src/test/scala/Null.scala
+++ b/src/sbt-test/sbt-scalafix/scalafixOnCompile/lint/src/test/scala/Null.scala
@@ -1,0 +1,3 @@
+object Null {
+  println(null)
+}
--- a/src/sbt-test/sbt-scalafix/scalafixOnCompile/project/plugins.sbt
+++ b/src/sbt-test/sbt-scalafix/scalafixOnCompile/project/plugins.sbt
@@ -1,0 +1,2 @@
+resolvers += Resolver.sonatypeRepo("public")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % sys.props("plugin.version"))
--- a/src/sbt-test/sbt-scalafix/scalafixOnCompile/rewrite/src/it/scala/UnusedImports.scala
+++ b/src/sbt-test/sbt-scalafix/scalafixOnCompile/rewrite/src/it/scala/UnusedImports.scala
@@ -1,0 +1,3 @@
+import java.time.Instant
+
+object UnusedImports
--- a/src/sbt-test/sbt-scalafix/scalafixOnCompile/rewrite/src/main/scala/UnusedImports.scala
+++ b/src/sbt-test/sbt-scalafix/scalafixOnCompile/rewrite/src/main/scala/UnusedImports.scala
@@ -1,0 +1,3 @@
+import java.time.Instant
+
+object UnusedImports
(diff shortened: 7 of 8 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -88,6 +96,29 @@ object ScalafixPlugin extends AutoPlugin {
...
+            val withOldCompile = Compat.append(
+              extracted,
+              Seq(compile.in(ref, config) := oldCompile),
+              currentState
+            )
+            extracted.runInputTask(
+              scalafix.in(ref, config),
+              input = "",
+              withOldCompile
+            )
```
bjaglin, line 118: This works, but it is very hacky, as the state "reload" is slow, and compilation runs for a given project/config are not de-duped during the liftetime of task execution... Still trying to figure out something better.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -88,6 +96,33 @@ object ScalafixPlugin extends AutoPlugin {
...
+            Seq(scalafix.key, scalafixAll.key).contains(root.key)
+          }
+          val runScalafixAfterCompile =
+            scalafixOnCompile.value && !scalafixRunExplicitly
+          val currentState = state.value
+          val ref = thisProjectRef.value
+          if (runScalafixAfterCompile) {
+            // We can't just run scalafix from here, as it would introduce a cyclic dependency
+            // since semantic rules require compilation products. Instead we run scalafix within
+            // a custom state where compile has not been overriden.
```
dwijnand, line 111: Where's the cycle?  `val oldCompile = compile.value` means "compile and store the result (`CompileAnalysis`) as `oldCompile`.  So scalafix should find all the compilation products it needs in the right places in the target directories.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
dwijnand: > 2\. compilations for a given project/config and most annoyingly all the projects it depends on are not de-duped during the liftetime of a task execution

Yes, because `runInputTask` is running outside of the task graph, which is dangerous for race conditions.  See these references for when sbt-buildinfo used to do the same thing, and the problems it caused: https://github.com/sbt/sbt-buildinfo/commit/7f3a548075b71b261cddc60fd6143ff7dfb87bf7#diff-0a369498a5a8db3ac8fa606b544c9810

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: on master: scalafix -> [fullClasspath](https://github.com/scalacenter/sbt-scalafix/blob/v0.9.18/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L346) -> compile
introduced in this PR: compile -> scalafix

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: Thanks for the pointers, that confirms my concerns! How did you get rid of it there? I am just wondering if it would be applicable here.

dwijnand: I changed the implementation so the values were defined within the Task (monad): https://github.com/sbt/sbt-buildinfo/compare/v0.7.0...v0.9.0#diff-60231996cd2e40cc694705deb78dc6d6R27; notice the `case BuildInfoKey.TaskValue`, which doesn't do the unsafe `extracted.runTask` (the `case BuildInfoKey.Task` is still there for backwards-compatibility and extreme cases.)

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
dwijnand: OK, instead of redefining `compile`, try something like:

```scala
scalafix := Def.taskDyn {
  // ...
  if (runScalafixAfterCompile)
    scalafix.triggeredBy(compile.in(config))
  else
    scalafix
}.value
```

bjaglin: I tried that earlier but as far as I remember, it did not give me what I wanted: a failure in `scalafix` woudn't cascade to `compile`, given the fire-and-forget semantics of `triggeredBy`.

dwijnand: I see, yes, that is to be expected.

So, thinking back, why does scalafix need fullClasspath but compile doesn't?  It looks like (ignoring a bunch of noise and detail) compile is defined in terms of `dependencyClasspath` and `classDirectory`, and `products` (a dependency of `fullClasspath`) is defined as 

```scala
  def makeProducts: Initialize[Task[Seq[File]]] = Def.task {
    compile.value
    copyResources.value
    classDirectory.value :: Nil
  }
```

So I wonder if your redefinition of compile could run old `compile` and `copyResources`, and then run `scalafix`, having redefined `scalafix` as depending on `dependencyClasspath` and `classDirectory`, the last of which should have the classfiles from `compile` and the resources from `copyResources`.

dwijnand: When `runScalafixAfterCompile == true` then `scalafix` would be an alias of `compile`, I guess.  Or something like that.

olafurpg:
Would it be possible to achieve similar functionality using `triggeredBy`? I'm couldn't get the following code to work as expected but I think the code looks something like this
```scala
    scalafix.in(Compile) :=
      scalafix.in(Compile).toTask(" ").triggeredBy(compile in Compile).value,
```

bjaglin:
> Would it be possible to achieve similar functionality using `triggeredBy`?

@olafurpg see inline discussion: compile wouldn't fail on lint error. I naively expected this behavior for visibility, but maybe it's better that way?

olafurpg:
My apologies! I missed that conversation, makes sense!

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: > having redefined scalafix as depending on dependencyClasspath and classDirectory, the last of which should have the classfiles from compile and the resources from copyResources.

By removing that dependency, the problem is that `scalafix` could run against stale class/semanticDB files, right?

bjaglin: I guess there is something to do with `scalafixRunExplicitly`, which can control in which direction the dependency is so it's never cyclic for a given task execution, I'll give it a try.

dwijnand: Another way to see it is: given scalafix will always depend on compilation, then "runScalafixAfterCompile" roughly means "redefine `compile` as `scalafix`".

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -32,17 +32,27 @@ object ScalafixPlugin extends AutoPlugin {
     val scalafix: InputKey[Unit] =
       inputKey[Unit](
         "Run scalafix rule(s) in this project and configuration. " +
-          "For example: scalafix RemoveUnusedImports. " +
-          "To run on test sources use test:scalafix or scalafixAll."
+          "For example: scalafix RemoveUnused. " +
+          "To run on test sources use test:scalafix or scalafixAll. " +
+          "When invoked directly, prior compilation will be triggered for semantic rules."
```
bjaglin, line 37: Not very proud of the "When invoked directly" wording but I am trying to get the attention of power users that would call/use scalafix input keys from another context (like another input meta input key that would aggregate scalafmt & scalafix for example). In that case, `scalafixRunExplicitly` will be false, so Scalafix might run on stale semanticdb/class files. I haven't found a way to work around that, but I think it's such a corner case and the current implementation so readable that it's not worth trying to address it.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: I think I came up with a reasonable implementation  It is a bit more tricky than I anticipated because the real `scalafix` can run any rule given that it's a input key, while the `scalafixOnCompile` can only trigger a no-arg `scalafix` which uses `.scalafix.conf` to find rules to run. That's why I changed a bit the behavior (see last paragraph of first comment).

Is that what you had in mind?

dwijnand: Yeah, looks good (only fast-scanned, this late at night 😄).

Nice to know it actually worked!  (famous last words...)

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -394,16 +414,22 @@ object ScalafixPlugin extends AutoPlugin {
         new SemanticdbNotFound(ruleNames, scalaVersion.value, sbtVersion.value)
       ).findErrors(files, dependencies, withScalaInterface)
       if (errors.isEmpty) {
-        Def.task {
+        val task = Def.task {
+          // passively consume compilation output without triggering compile as it can result in a cyclic dependency
+          val classpath =
+            dependencyClasspath.in(config).value.map(_.data.toPath) :+
+              classDirectory.in(config).value.toPath
```
bjaglin, line 421: I have left out `[copyResources`](https://github.com/sbt/sbt/blob/v1.3.13/main/src/main/scala/sbt/Defaults.scala#L3125) since I don't see why Scalafix would need it but it might be a wrong assumption.

bjaglin:
I can't think of any downside redefining `compile` when `scalafixOnCompile := false`, except maybe polluting a tiny bit the `inspect tree` for `compile`?

As for what happens when `scalafixOnCompile := true`, I am not sure the order in which plugins are (explicitly) enabled matter, and therefore if it makes any difference to have a plugin that does not trigger automatically so that users can control ordering with other compile-hijacking plugins. sbt-scalafmt runs [before compilation](https://github.com/scalameta/sbt-scalafmt/blob/23e308f8daa118153023fc9f0dec7a50d84117fe/plugin/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala#L402-L404), so that particular plugin is not a problem.

I'd tend to stick to the current implementation for simplicity unless we have factual elements that demonstrates its shortcomings.

@dwijnand, since you are here, your thoughts on this are more than welcome!

bjaglin:
@olafurpg any thoughts on https://github.com/scalacenter/sbt-scalafix/pull/140/commits/a4ba4964420fac47d8633a6b68d4779e73b39708 in particular?

dwijnand:
I think Sébastien said it best (somewhere): if a plugin changes the default build settings/tasks, then it should be opt-in enabled, not auto-trigger from being on the classpath.  But it's not a practice that is even documented or even clearly, expressly understood between major sbt plugin authors and sbt maintainers, so it's clearly not strongly observed by plugins in the ecosystem.

So, given that, I don't have any immediate feedback to give (reason why I just +1'd).  But, eventually, it be better to make the plugin opt-in.

I don't even know what the behaviour would be if you did `enablePlugins(ScalafmtOnCompilePlugin, ScalafixOnCompilePlugin)` (let's say the "before compilation" thing weren't true).  It's probably either order of definition or alphabetic ordering, but it could even be something random-looking like the order of hashing something.  I know that when sbt's `Plugin` were re-designed with `AutoPlugin` the focus was on the plugin dependencies part, in particular for SbtWeb and SbtPlay: before you could get wrong behaviour if you didn't enable correctly, afterwards the order didn't matter and triggers happened as expected.  But (as I said) I don't know what the behaviour is for unrelated (technically) plugins like Scalafmt and Scalafix, and I doubt it's documented.

bjaglin:
Thanks a lot for the context @dwijnand !

Since we are pre 1.0 (which maybe wasn't the case for `scalafmt` that took the `scalafmtOnCompile` road?), I guess we should go for the separate plugin way from the start.

@olafurpg should I merge this as-is and you open a PR with your extra commits?

olafurpg:
Pinging @sjrd to weigh in on the discussion starting at https://github.com/scalacenter/sbt-scalafix/pull/140#pullrequestreview-446231603

Do you think that it's acceptable/idiomatic to redefine the `compile` task by default as is done in the current PR or do you think it would be preferable to move this logic into a separate off-by-default plugin? I wish there was a more lightweight syntax to enable an off-by-default plugin than creating `project/MyPlugin.scala`, from the user's perspective it's a lot of additional ceremony compared to `ThisBuild / scalafixOnCompile := true`.

olafurpg:
@bjaglin It's a good point that sbt-scalafmt formats before compile so there shouldn't be an integration issue. However, I suppose users would ideally prefer Scalafmt to format after Scalafix when you enable scalafix+scalafmt on compile. Pinging in @poslegm who has worked on the sbt-scalafmt plugin, I wonder if it's worth considering changing the sbt-scalafmt plugin to make it possible to format after Scalafix completes?

sjrd:
@olafurpg: @dwijnand explained my position quite well with regards to modifying existing tasks and settings by default. In general, if you're modifying a task or setting that is not defined by your plugin, there's one big no no: do not make your plugin auto-triggered. Because of how plugins get automatically loaded simply because of being on the classpath (including through transitive dependencies), having auto-triggered plugins that modify existing stuff can cause surprising behaviors, sometimes difficult or impossible to disable.

That said, the way that `compile` is overridden here is very well done. If the config is such that fix-on-compile is not true (and that being the default), `compile` basically ends up being overridden by itself, which basically amounts to doing nothing. The rule above about not changing others' settings and tasks in an auto-triggered plugin does not apply.

Another big danger is that any `taskDyn` or `settingDyn` worsens a) the output of `inspect tree`, potentially hiding large swathes of depended on keys and b) the ability for sbt to parallelize task execution, as it needs to know the result of the dynamism before proceeding with the execution of the dynamic aspects. Again, the way that this PR overrides `compile` means, I believe, that the parallelism of `compile` will be essentially unaffected. However, scalafixing of course runs sequentially, but that's expected.

So overall, from a technical standpoint of overriding `compile`, I think this PR is sound and does this the correct way.

---

Now, you didn't ask my opinion, but I can't comment on a thread like this without giving it anyway: modifying files in a task (either input files or files produced by earlier tasks) is a very bad thing, and you should never do that, ever. The parallelism of sbt means that things can execute concurrently on already created files, causing very bad issues. If you want to modify things, use a *command* instead (and no, you can't execute a command from a task, not in a sound way anyway).

OK, I said it once, I will now hold my peace about this.

bjaglin:
> modifying files in a task (either input files or files produced by earlier tasks) is a very bad thing, and you should never do that, ever

Good point, which I guess goes beyond the scope of this PR as `scalafix` is already an (input) task mutating input files. However, triggering it from `compile` would probably expose the problem much more.

Since the `scalafix` input task's purpose is to transform `unmanagedSources` output, would it be possible to prevent races by tagging `unmanagedSources` and `scalafix` with a custom tag, limiting one concurrent execution by project/config?

bjaglin:
> Since the `scalafix` input task's purpose is to transform [`unmanagedSources` output](https://github.com/scalacenter/sbt-scalafix/blob/66b5f297811ee60796c51967867d507b42e6f349/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L558), would it be possible to prevent races by tagging `unmanagedSources` and `scalafix` with a custom tag, enforcing at most one concurrent execution by project/config?

nevermind, that wouldn't prevent tasks depending on `unmanagedSources` to run concurrently with `scalafix`....

bjaglin:
I guess we could still use `Tags.customLimit` to enforce that `scalafix` does not run concurrently with any other task from the same project/config. To be completely correct, it shouldn't run concurrently with anything at all (`Tags.exclusive`) since sources might be read across projects/configs, but that would really impact uncached `compile` invocations on multi-projects builds with `scalafixOnCompile := true`.

poslegm:
> Pinging in @poslegm who has worked on the sbt-scalafmt plugin, I wonder if it's worth considering changing the sbt-scalafmt plugin to make it possible to format after Scalafix completes?

I think it's OK to make changes in sbt-scalafmt because these are projects from the same ecosystem. But I can't figure out yet exactly how to track the scalafix finish from another plugin.

olafurpg:
Thank you @sjrd for pitching in! That was helpful. With all things considered, I think it's OK to merge this PR as is. I'm not convinced it's worth complicating the installation setup with a separate `ScalafixOnCompilePlugin`.

I agree with you that modifying files during compile is a bad thing. I have tried to prevent people from running formatting/fixing on compile since 2016 with little luck so far 😅 As for modifying files from any task I agree in principle but I haven't yet figured out how to implement it in practice so far 🤷

olafurpg:
@poslegm It's a good question. I feel like it might be possible to achieve this by defining a callback hook in sbt-scalafix that sbt-scalafmt can register.

```scala
// sbt-scalafix

val scalafixOnCompileHook = taskKey[Seq[File] => Unit]("Callback hook after Scalafix has completed fixing a file")
scalafix := {
  val filesToFix = // .. command like usual
  scalafixOnComipileHook.?.value.foreach(hook => hook(filesToFix))
}

// sbt-scalafmt

TaskKey[Seq[File] => Unit]("scalafixOnCompileHook") := { files => formatFiles(files) }
```

You can achieve something similar by introducing a direct dependency between the plugins but I don't think that's necessary.

Either way, feel free to ignore these thoughts. This is not blocking this PR, I'm just thinking out loud.

bjaglin:
> I feel like it might be possible to achieve this by defining a callback hook in sbt-scalafix that sbt-scalafmt can register.

:+1: let's follow-up on implementation details in https://github.com/scalacenter/scalafix/issues/1206

eed3si9n:
I'm somewhat wary of `compile` (most called task in sbt) turning into a dynamic task:

```scala
        compile := Def.taskDyn {
          val oldCompile =
            compile.value // evaluated first, before the potential scalafix evaluation
          val runScalafixAfterCompile =
            scalafixOnCompile.value && !scalafixRunExplicitly.value
          if (runScalafixAfterCompile)
            scalafix
              .toTask("")
              .map(_ => oldCompile)
          else Def.task(oldCompile)
        }.value,
```

When Play needed post processing, Josh and I split `compile` task into `compileIncremental` and `manipulateBytecode` task, which plugins can hook into. I wonder if Scalafix can use that hook? By default:

```scala
    manipulateBytecode := compileIncremental.value,
```

but you could probably do:

```scala
Compile / manipulateBytecode := {
  val orig = (Compile / manipulateBytecode).value // respect the other plugins
  println("post processing")
  orig
}
```

bjaglin:
Thanks for chiming-in @eed3si9n!

> I'm somewhat wary of compile (most called task in sbt) turning into a dynamic task:

Could you explain why it's better to redefine (with a dynamic wrapper task ) a subtask of compile rathe than `compile` itself? I can understand it "blurs out" less the task graph (`inspect tree`), but is there any runtime impact as well?

https://github.com/bjaglin/sbt-scalafix/commit/370219524a0311898254711e9ef8a263a50f3f86 seems to work just fine, but I didn't manage to write a proper motivation in the commit message, thus my question.

eed3si9n:
@bjaglin I don't think there will be noticeable performance impact.

One potential benefit of hooking into `manipulateBytecode` is that it's still in the `compile` change of task, which means the rest of the `compileTask` (https://github.com/sbt/sbt/blob/v1.3.13/main/src/main/scala/sbt/Defaults.scala#L1716-L1735) won't happen on failure, which includes writing Analysis file into the disk. After all you're making `compile` task fail, so that should correspond with the Analysis.

olafurpg:
> writing Analysis file into the disk. After all you're making compile task fail, so that should correspond with the Analysis.

In the case of Scalafix, I think it would be desirable to persist incremental compile analysis files even in case Scalafix fails. As a user, I would be surprised if a Scalafix linting error would cause sbt to re-compile the Scala sources.

sjrd:
Random thought: have you considered using a `triggeredBy(compile)` mechanism, instead of modifying `compile`? This should avoid technical concerns about performance/correctness of modifying `compile`.

bjaglin:
> Random thought: have you considered using a `triggeredBy(compile)` mechanism, instead of modifying `compile`?

[We have](https://github.com/scalacenter/sbt-scalafix/pull/140#issuecomment-654323814): it wouldn't fail verbosely in case of a lint error.

bjaglin:
I think I will go ahead with this as-is.

#### Outstanding comments

https://github.com/scalacenter/sbt-scalafix/pull/140#issuecomment-656663541
> modifying files in a task (either input files or files produced by earlier tasks) is a very bad thing, and you should never do that, ever. The parallelism of sbt means that things can execute concurrently on already created files, causing very bad issues.

I'll follow-up with a PR with a tentative fix for that based on https://github.com/scalacenter/sbt-scalafix/pull/140#issuecomment-656692574.

---
https://github.com/scalacenter/sbt-scalafix/pull/140#issuecomment-656721386
> I'm somewhat wary of compile (most called task in sbt) turning into a dynamic task
> When Play needed post processing, Josh and I split compile task into compileIncremental and manipulateBytecode task, which plugins can hook into. I wonder if Scalafix can use that hook?

Given the [change that this brings](https://github.com/scalacenter/sbt-scalafix/pull/140#issuecomment-657013111), I feel like it's still better to override `compile`.

---

https://github.com/scalacenter/sbt-scalafix/pull/140#issuecomment-656699864
> I feel like it might be possible to achieve this by defining a callback hook in sbt-scalafix that sbt-scalafmt can register.

Will be handled separately, see https://github.com/scalacenter/scalafix/issues/1206.

olafurpg:
I agree @bjaglin. LGTM 🚀

PR #141, opened 2020-07-06
prevent failure on sbt init when JGit finds a .git with no commit

bjaglin:
Ref https://github.com/scalacenter/scalafix/issues/1196
Regression detected between 0.9.17 & 0.9.18, introduced by https://github.com/scalacenter/sbt-scalafix/pull/126

2ca68e4 exposed the failure to lookup commits during sbt initialization, while it would only happen on `scalafix` invocation earlier.

While looking at that, I initially thought 2ca68e4 changed the behavior by eagerly fetching the last commits into a setting, but [that lookup is memoized in a lazy val anyway](https://github.com/scalacenter/sbt-scalafix/blob/v0.9.17/src/main/scala/scalafix/internal/sbt/ScalafixCompletions.scala#L98-L99), so a `sbt reload` is required before or after 2ca68e4 to get an up-to-date list of commits.

### DIFF
--- a/src/test/scala/scalafix/internal/sbt/JGitCompletionsSuite.scala
+++ b/src/test/scala/scalafix/internal/sbt/JGitCompletionsSuite.scala
@@ -1,0 +1,28 @@
+package scalafix.internal.sbt
+
+import org.eclipse.jgit.api.{Git => JGit}
+import org.scalatest.funsuite.AnyFunSuite
+
+class JGitCompletionsSuite extends AnyFunSuite {
+
+  test("directory with no .git") {
+    val fs = new Fs()
+    val jgit = new JGitCompletion(fs.workingDirectory)
+    assert(jgit.last20Commits == Nil)
+  }
+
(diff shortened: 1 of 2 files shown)


src/test/scala/scalafix/internal/sbt/JGitCompletionsSuite.scala
```diff
@@ -0,0 +1,27 @@
...
+    val jgit = new JGitCompletion(fs.workingDirectory)
+    assert(jgit.last20Commits == Nil)
+  }
+
+  test("directory with no commits") {
+    val fs = new Fs()
+    new Git(fs.workingDirectory) // git init
+    val jgit = new JGitCompletion(fs.workingDirectory)
+    assert(jgit.last20Commits == Nil)
+  }
```
bjaglin, line 25: that's the one reproducing https://github.com/scalacenter/scalafix/issues/1196#issuecomment-654153643 (the other tests pass on master, while this one fails)

PR #142, opened 2020-07-06
Fix 0.9.18 caching/performance regressions

bjaglin:
* The first commit fixes an embarrassingly massive performance regression introduced by 2ca68e4: the comment/assumption in https://github.com/scalacenter/sbt-scalafix/pull/126#discussion_r439692348 was wrong, causing artifact resolution & classloader creation to happen for each invocation/project/configuration, which caused slow startup and slower rule execution (JIT not warmed up). It might also fix https://github.com/scalacenter/scalafix/issues/1196#issuecomment-654744979.
* The second commit brings back the 0.9.17 behavior of looking up last commits for the parser at the first `scalafix` invocation (changed also in 2ca68e4), to reduce the risk of JGit stuff blocking the initialization like https://github.com/scalacenter/scalafix/issues/1196. Related to https://github.com/scalacenter/sbt-scalafix/pull/141.
* The two last commits fix a minor known regression introduced by https://github.com/scalacenter/sbt-scalafix/commit/d81dd7266231e70af475551b563b21ec75c89482 (see message of that commit).

We'll probably need some performance tests to catch those regressions in the future...

### Testing on a real project

I have timed the initialization overhead of `scalafix` by running a warm `all scalafix test:scalafix it:scalafix` (third invocation with `scalafixCaching := true`) on a build with 25 sub-projects which have Scalafix enabled.

* With `scalafixDependencies`
  * 0.9.16: ~10s
  * 0.9.18: ~50s
  * This PR: ~8s

* With `scalafixDependencies` & `dependsOn(customRules % ScalafixConfig)`
  * 0.9.16: ~10s
  * 0.9.18: ~50s
  * This PR: ~9s

The impact of (not) JITing the rules is not exposed by this test, but rule execution should be back to pre-0.9.18 level as well.

### DIFF
--- a/src/main/scala/scalafix/internal/sbt/BlockingCache.scala
+++ b/src/main/scala/scalafix/internal/sbt/BlockingCache.scala
@@ -1,0 +1,21 @@
+package scalafix.internal.sbt
+
+import scala.collection.mutable
+
+/** A basic thread-safe cache without any eviction. */
+class BlockingCache[K, V] {
+  private val underlying = new mutable.HashMap[K, V]
+
+  /**
+    * @param value By-name parameter evaluated when the key if missing. Value computation is guaranteed
+    *              to be called only once per key across all invocations.
+    */
+  def getOrElseUpdate(key: K, value: => V): V = {
+    // ConcurrentHashMap does not guarantee that there is only one evaluation of the value, so
+    // we use our own (global) locking, which is OK as the number of keys is expected to be
+    // very small (bound by the number of projects in the sbt build).
+    underlying.synchronized {
+      underlying.getOrElseUpdate(key, value)
+    }
+  }
+}
--- a/src/main/scala/scalafix/internal/sbt/BlockingCache.scala
+++ b/src/main/scala/scalafix/internal/sbt/BlockingCache.scala
@@ -1,21 +1,22 @@
 package scalafix.internal.sbt
 
-import scala.collection.mutable
+import java.{util => ju}
 
 /** A basic thread-safe cache without any eviction. */
 class BlockingCache[K, V] {
-  private val underlying = new mutable.HashMap[K, V]
+
+  // Number of keys is expected to be very small so the global lock should not be a bottleneck
+  private val underlying = ju.Collections.synchronizedMap(new ju.HashMap[K, V])
 
   /**
     * @param value By-name parameter evaluated when the key if missing. Value computation is guaranteed
     *              to be called only once per key across all invocations.
     */
-  def getOrElseUpdate(key: K, value: => V): V = {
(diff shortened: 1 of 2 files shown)


src/main/scala/scalafix/internal/sbt/BlockingCache.scala
```diff
@@ -0,0 +1,21 @@
+package scalafix.internal.sbt
+
+import scala.collection.mutable
+
+/** A basic thread-safe cache without any eviction. */
+class BlockingCache[K, V] {
```
olafurpg, line 6: FWIW, if you don't mind using Java collections you can replace this class with
```scala
import java.{util => ju}
object BlockingCache {
  def empty[A, B] =
    ju.Collections.synchronizedMap(new ju.HashMap[A, B])
}

val cache = BlockingCache.empty[Int, String]
cache.computeIfAbsent(10, _ => "9")

cache // {10=9}
```
Those APIs are quite quirky however (return `AnyRef` in many places) so I like rolling a custom Scala solution instead.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -198,18 +219,27 @@ object ScalafixPlugin extends AutoPlugin {
...
-      if (refreshClasspath)
-        scalafixInterface().withArgs(
-          ToolClasspath(
-            projectDepsInternal.map(_.toURI.toURL),
-            baseDepsExternal ++ projectDepsExternal ++ rulesDepsExternal,
-            baseResolvers
-          )
+      if (customToolClasspath) {
+        val toolClasspath = ToolClasspath(
+          projectDepsInternal0.map(_.toURI.toURL),
```
olafurpg, line 233: Unrelated to this PR, I see that we're using `java.net.URL` as cache keys. This is problematic because `URL.equals` does remote blocking calls

> Two hosts are considered equivalent if both host names can be resolved into the same IP addresses; else if either host name can't be resolved, the host names must be equal without regard to case; or both host names equal to null.

https://docs.oracle.com/javase/6/docs/api/java/net/URL.html#equals(java.lang.Object)

It's probably better to use `URI` instead of `URL`

src/main/scala/scalafix/internal/sbt/BlockingCache.scala
bjaglin: good idea - I kept the facade but swapped the implementation in e822f41

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: :+1: already got bitten by that, but apparently not enough to care... will change that in a separate PR

PR #144, opened 2020-07-07
scalafixAll tweaks

bjaglin:
See commit messages


bjaglin:
@olafurpg I was inspired by https://github.com/sbt/sbt/pull/5023/files for the task naming. It was very confusing to me to see the same task name appear several times in supershell (especially after I fixed the duplication bug from the first commit).

PR #145, opened 2020-07-10
caching: remove optimization bringing false positive cache hits

bjaglin:
https://github.com/scalacenter/scalafix/issues/1164#issuecomment-656567498

If a file is patched during a scalafix run, it's not safe to assume that all rules observed the updated content of that file and thus can
be skipped next time. It is if we have only one rule (which we could check) or if the patch is applied after the first rule (which we can't tell). For correctness, stamping should happen before rule execution.

### DIFF
--- a/src/sbt-test/skip-windows/caching/test
+++ b/src/sbt-test/skip-windows/caching/test
@@ -34,18 +34,17 @@ $ exec chmod 000 src/main/scala/Valid.scala
 > scalafix --check
 $ delete src/main/scala
 
-# a file patched during the last run should be cached
+# a file patched during the last run should not be cached as some rules might not have observed the latest content
 > set scalafixConfig := None
 $ mkdir src/test/scala
 $ copy-file files/ProcedureSyntax.scala src/test/scala/ToPatch.scala
 $ copy-file files/Valid.scala src/test/scala/Valid.scala
 # avoid rounding in mtime that could cause false negatives in `newer`
 $ sleep 1000
-> test:scalafix ProcedureSyntax
+> test:scalafix ProcedureSyntax DisableSyntax
 $ newer src/test/scala/ToPatch.scala src/test/scala/Valid.scala
 $ exec chmod 000 src/test/scala/ToPatch.scala
-$ exec chmod 000 src/test/scala/Valid.scala
-> test:scalafix ProcedureSyntax
+-> test:scalafix ProcedureSyntax DisableSyntax
 $ delete src/test/scala
 
 # an added file after a successful run should be checked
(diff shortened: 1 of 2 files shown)


src/sbt-test/skip-windows/caching/test
```diff
@@ -216,6 +217,7 @@ $ exec chmod 000 src/main/scala/Valid.scala
 > scalafix --check ProcedureSyntax
 -> test:scalafix --check ProcedureSyntax
 > test:scalafix ProcedureSyntax
+> test:scalafix ProcedureSyntax
```
bjaglin, line 220: a second warm is needed for the cache to converge since there is a rewrite involved

src/sbt-test/skip-windows/caching/test
```diff
@@ -34,18 +34,19 @@ $ exec chmod 000 src/main/scala/Valid.scala
...
-# a file patched during the last run should be cached
+# a file patched during the last run should not be cached as some rules might not have observed the latest content
 > set scalafixConfig := None
 $ mkdir src/test/scala
 $ copy-file files/ProcedureSyntax.scala src/test/scala/ToPatch.scala
 $ copy-file files/Valid.scala src/test/scala/Valid.scala
 # avoid rounding in mtime that could cause false negatives in `newer`
 $ sleep 1000
-> test:scalafix ProcedureSyntax
+> test:scalafix ProcedureSyntax DisableSyntax
```
bjaglin, line 44: adding a second rule for exhibiting the scenario, even though behavior is the same no matter how many rules we have (but it might change one day)

src/sbt-test/skip-windows/caching/test
```diff
@@ -34,18 +34,19 @@ $ exec chmod 000 src/main/scala/Valid.scala
...
 $ newer src/test/scala/ToPatch.scala src/test/scala/Valid.scala
-$ exec chmod 000 src/test/scala/ToPatch.scala
 $ exec chmod 000 src/test/scala/Valid.scala
-> test:scalafix ProcedureSyntax
+-> test:scalafix ProcedureSyntax DisableSyntax
+$ exec chmod 000 src/test/scala/ToPatch.scala
+> test:scalafix ProcedureSyntax DisableSyntax
```
bjaglin, line 49: verify the cache remains incremental for non-patched files

src/sbt-test/skip-windows/caching/test
```diff
@@ -34,18 +34,17 @@ $ exec chmod 000 src/main/scala/Valid.scala
...
 $ mkdir src/test/scala
 $ copy-file files/ProcedureSyntax.scala src/test/scala/ToPatch.scala
 $ copy-file files/Valid.scala src/test/scala/Valid.scala
 # avoid rounding in mtime that could cause false negatives in `newer`
 $ sleep 1000
-> test:scalafix ProcedureSyntax
+> test:scalafix ProcedureSyntax DisableSyntax
 $ newer src/test/scala/ToPatch.scala src/test/scala/Valid.scala
 $ exec chmod 000 src/test/scala/ToPatch.scala
-$ exec chmod 000 src/test/scala/Valid.scala
```
bjaglin, line 47: irrelevant to the assertion, it's just useful to assert above via `newer` that `ToPatch` was updated

PR #147, opened 2020-07-10
add setting key to register callbacks after Scalafix runs

bjaglin:
First step of https://github.com/scalacenter/scalafix/issues/1206

### DIFF
--- a/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
+++ b/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
@@ -10,6 +10,7 @@ import scalafix.interfaces.{Scalafix => ScalafixAPI, _}
 import scalafix.sbt.InvalidArgument
 
 import scala.collection.JavaConverters._
+import scala.util.Try
 import scala.util.control.NonFatal
 
 sealed trait Arg extends (ScalafixArguments => ScalafixArguments)
@@ -97,6 +98,7 @@ object Arg {
 
 class ScalafixInterface private (
     scalafixArguments: ScalafixArguments, // hide it to force usage of withArgs so we can intercept arguments
+    postScalafixHooks: Seq[Seq[File] => Unit],
     val args: Seq[Arg]
 ) {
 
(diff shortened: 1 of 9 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -115,6 +115,13 @@ object ScalafixPlugin extends AutoPlugin {
 
   import autoImport._
 
+  private val postScalafixHooks: SettingKey[Seq[Seq[File] => Unit]] =
```
bjaglin, line 118: I guess we want to keep this hidden for now? I have used plural so that we can safely advertise it to other plugins in the feature without breaking sbt-scalafmt

src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
```diff
@@ -105,11 +107,33 @@ class ScalafixInterface private (
...
+      postScalafixHooks,
+      this.args ++ args
+    )
   }
 
   def run(): Seq[ScalafixError] =
-    scalafixArguments.run().toSeq
+    try {
+      scalafixArguments.run().toSeq
+    } finally {
```
bjaglin, line 120: it might be possible to be more precise (run this only on `LinterError`), but I decided to be safe and trigger the callback whenever this runs, because mutations are applied after each rule so there is no guarantee that nothing was touched when we get back an exception

src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
```diff
@@ -105,11 +107,33 @@ class ScalafixInterface private (
...
+          Seq("--check", "--test", "--auto-suppress-linter-errors")
+            .intersect(args)
+            .isEmpty
+        }
+      if (mayRewrite) for {
+        hook <- postScalafixHooks
+        files <- args.reverse.collectFirst {
+          case Arg.Paths(paths) => paths.map(_.toFile)
+        }
+      } yield Try(hook(files))
```
bjaglin, line 135: after implementing this, I am wondering if it's really relevant to pass which files might have changed, as scalafmt has its own targets and its own cache.

src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
```diff
@@ -105,11 +107,33 @@ class ScalafixInterface private (
...
+    } finally {
+      val mayRewrite = args.reverse
+        .collectFirst {
+          case Arg.ParsedArgs(args) => args
+        }
+        .forall { args =>
+          Seq("--check", "--test", "--auto-suppress-linter-errors")
+            .intersect(args)
+            .isEmpty
+        }
```
bjaglin, line 129: that's the only solid heuristic I could think of to reduce the number of unnecessary callbacks

bjaglin:
I am closing this for now as I am now convinced it's not the way to go. I'll try to add a `scalafixPostCompileRun` task key on top of https://github.com/scalacenter/sbt-scalafix/pull/140 that `sbt-scalafmt` can declare itself and redefine (if it exists), to postpone formatting. That won't give information about which files might have changed, but this information was not really useful anyway.

olafurpg:
Sounds good. Reviewing the code I agree. I think it's also worth exploring another approach where we have a scalafmt rule that formats the code and we extend `.scalafmt.conf` to support running scalafix in several independent steps. This shouldn't be too hard to support when combined with https://github.com/scalacenter/scalafix/issues/1204

PR #148, opened 2020-07-13
add scalafixOnCompileConfig to separate .scalafix.conf for scalafix-on-compilation

taisukeoe:
While I'm playing round scalafixOnCompile #140 now, I find a problem that may disturb current scalafix users to use RemoveUnused or OrganizeImports.

## Problem
Some of popular Scalafix rules like RemoveUnused or OrganizeImports work terribly with `scalafixOnCompile := true`.

Given the situation that `scalafixOnCompile` is true and RemoveUnused or OrganizeImports rules are effective,
writing an import statement and compiling it before adding a user code of the import, and then the import disappears.

Same problem exists in private or local variable with RemoveUnused rule.

Generally speaking, I believe Scalafix is great because its rules can vary in terms of their use-cases. Introducing `scalafix-on-compile` sounds great, because it might expand scalafix use-cases and can be an enabler to use scalafix as an extra linter like WartRemover.

Some rules can work great with scalafix-on-compilation already while there must be ones which are designed only for scalafix-explicit-run.

## Solution in this PR
So I'd think there must be a way to separate Scalafix rules between running explicitly and running on compile.

In this PR, I introduce scalafixConfingOnCompile settingKey to specify alternative `.scalafix.conf` which will be used only in `onCompile`.
At default, scalafixConfigOnCompile points to scalafixConfig so it won't affect users who don't need alternative `.scalafix.conf`.

If there is better way to solve the problem, please feel free to let me know. Your feedback is welcome.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -89,7 +89,7 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional location to .scalafix.conf file to specify which scalafix rules should run. " +
           "Defaults to the build base directory if a .scalafix.conf file exists."
       )
-    val scalafixConfigOnCompile: SettingKey[Option[File]] =
+    val scalafixOnCompileConfig: SettingKey[Option[File]] =
       settingKey[Option[File]](
         "Optional location to .scalafix.conf file to specify which scalafix rules should run on compile. " +
           "Defaults to the same as scalafixConfig."
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -91,8 +91,8 @@ object ScalafixPlugin extends AutoPlugin {
       )
     val scalafixOnCompileConfig: SettingKey[File] =
       settingKey[File](
-        "Optional location to .scalafix.conf file to specify which scalafix rules should run on compile. " +
-          "Defaults to the same as scalafixConfig."
+        "Location to .scalafix.conf file to specify which scalafix rules should run on compile. " +
+          "scalafixOnCompile must be set for this to have an effect. Defaults to the same as scalafixConfig."
       )
     val scalafixSemanticdb: ModuleID =
       scalafixSemanticdb(BuildInfo.scalametaVersion)
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -89,6 +89,11 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional location to .scalafix.conf file to specify which scalafix rules should run. " +
           "Defaults to the build base directory if a .scalafix.conf file exists."
       )
+    val scalafixConfigOnCompile: SettingKey[Option[File]] =
+      settingKey[Option[File]](
+        "Optional location to .scalafix.conf file to specify which scalafix rules should run on compile. " +
+          "Defaults to the same as scalafixConfig."
+      )
     val scalafixSemanticdb: ModuleID =
       scalafixSemanticdb(BuildInfo.scalametaVersion)
     def scalafixSemanticdb(scalametaVersion: String): ModuleID =
@@ -180,6 +185,7 @@ object ScalafixPlugin extends AutoPlugin {
 
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     scalafixConfig := None, // let scalafix-cli try to infer $CWD/.scalafix.conf
+    scalafixConfigOnCompile := scalafixConfig.value,
     scalafixOnCompile := false,
     scalafixResolvers := Seq(
       Repository.ivy2Local(),
@@ -344,7 +350,9 @@ object ScalafixPlugin extends AutoPlugin {
       if (shellArgs.rules.isEmpty && shellArgs.extra == List("--help")) {
         scalafixHelp
       } else {
-        val scalafixConf = scalafixConfig.in(config).value.map(_.toPath)
+        val scalafixConf =
+          if (scalafixRunExplicitly.value) scalafixConfig.in(config).value
+          else scalafixConfigOnCompile.in(config).value
         val (shell, mainInterface0) = scalafixArgsFromShell(
           shellArgs,
           scalafixInterfaceProvider.value,
@@ -361,7 +369,7 @@ object ScalafixPlugin extends AutoPlugin {
           .withArgs(maybeNoCache: _*)
           .withArgs(
             Arg.PrintStream(errorLogger),
-            Arg.Config(scalafixConf),
+            Arg.Config(scalafixConf.map(_.toPath)),
             Arg.Rules(shell.rules),
             Arg.ParsedArgs(shell.extra)
           )
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -50,6 +50,12 @@ object ScalafixPlugin extends AutoPlugin {
           "but can be disabled with `scalafixCaching := false`."
       )
 
+    val scalafixOnCompileCheck: InputKey[Unit] =
+      inputKey[Unit](
+        "Run `scalafix --check` against rule(s) declared in scalafixOnCompileConfig." +
+          "If scalafixOnCompileConfig is undefined, run checking against scalafixConfig or .scalafix.conf ."
+      )
+
     val scalafixCaching: SettingKey[Boolean] =
       settingKey[Boolean](
         "Cache scalafix invocations (off by default, on if scalafixOnCompile := true)."
@@ -102,6 +108,7 @@ object ScalafixPlugin extends AutoPlugin {
     def scalafixConfigSettings(config: Configuration): Seq[Def.Setting[_]] =
       Seq(
         scalafix := scalafixInputTask(config).evaluated,
+        scalafixOnCompileCheck := scalafixOnCompileInputTask(config).evaluated,
         compile := Def.taskDyn {
           val oldCompile =
             compile.value // evaluated first, before the potential scalafix evaluation
@@ -316,6 +323,29 @@ object ScalafixPlugin extends AutoPlugin {
         }
         .joinWith(_.join.map(_ => ()))
     }
+
+  private def scalafixOnCompileInputTask(
+      config: Configuration
+  ): Def.Initialize[InputTask[Unit]] =
+    // workaround https://github.com/sbt/sbt/issues/3572 by invoking directly what Def.inputTaskDyn would via macro
+    InputTask
+      .createDyn(InputTask.initParserAsInput(scalafixCompletions(_.parser)))(
+        Def.task { shellArgs =>
+          val actualOnCompileConf = scalafixOnCompileConfig
+            .in(config)
+            .?
+            .value
+            .orElse(scalafixConfig.in(config).value)
+
+          val overrideArgs = ShellArgs(extra =
+            "--check" ::
+              actualOnCompileConf.toList.flatMap(c =>
+                List("--config", c.getAbsolutePath)
+              )
+          )
+          scalafixTask(shellArgs + overrideArgs, config)
+        }
+      )
 
   private def scalafixInputTask(
       config: Configuration
@@ -581,7 +611,9 @@ object ScalafixPlugin extends AutoPlugin {
   private val scalafixRunExplicitly: Def.Initialize[Task[Boolean]] =
     Def.task {
       executionRoots.value.exists { root =>
-        Seq(scalafix.key, scalafixAll.key).contains(root.key)
+        Seq(scalafix.key, scalafixAll.key, scalafixOnCompileCheck.key).contains(
+          root.key
+        )
       }
     }
 
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -116,7 +116,15 @@ object ScalafixPlugin extends AutoPlugin {
             scalafixOnCompile.value && !scalafixRunExplicitly.value
           if (runScalafixAfterCompile)
             scalafix
-              .toTask("")
+              .toTask {
+                // Since last scalafix.conf wins, scalafixOnCompileConfig is passed if it's defined.
+                // Otherwise no argument is passed here, and scalafixConfig or .scalafix.conf will be used.
+                scalafixOnCompileConfig
+                  .in(config)
+                  .?
+                  .value
+                  .fold("")(conf => s" --config=${conf.getAbsolutePath}")
+              }
               .map(_ => oldCompile)
           else Def.task(oldCompile)
         }.value,
@@ -379,14 +387,7 @@ object ScalafixPlugin extends AutoPlugin {
       if (shellArgs.rules.isEmpty && shellArgs.extra == List("--help")) {
         scalafixHelp
       } else {
-        val scalafixConf =
-          if (scalafixRunExplicitly.value) scalafixConfig.in(config).value
-          else
-            scalafixOnCompileConfig
-              .in(config)
-              .?
-              .value
-              .orElse(scalafixConfig.in(config).value)
+        val scalafixConf = scalafixConfig.in(config).value
         val (shell, mainInterface0) = scalafixArgsFromShell(
           shellArgs,
           scalafixInterfaceProvider.value,
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -89,8 +89,8 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional location to .scalafix.conf file to specify which scalafix rules should run. " +
           "Defaults to the build base directory if a .scalafix.conf file exists."
       )
-    val scalafixOnCompileConfig: SettingKey[Option[File]] =
-      settingKey[Option[File]](
+    val scalafixOnCompileConfig: SettingKey[File] =
+      settingKey[File](
         "Optional location to .scalafix.conf file to specify which scalafix rules should run on compile. " +
           "Defaults to the same as scalafixConfig."
       )
@@ -185,7 +185,6 @@ object ScalafixPlugin extends AutoPlugin {
 
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     scalafixConfig := None, // let scalafix-cli try to infer $CWD/.scalafix.conf
-    scalafixConfigOnCompile := scalafixConfig.value,
     scalafixOnCompile := false,
     scalafixResolvers := Seq(
       Repository.ivy2Local(),
@@ -352,7 +351,12 @@ object ScalafixPlugin extends AutoPlugin {
       } else {
         val scalafixConf =
           if (scalafixRunExplicitly.value) scalafixConfig.in(config).value
-          else scalafixConfigOnCompile.in(config).value
+          else
+            scalafixOnCompileConfig
+              .in(config)
+              .?
+              .value
+              .orElse(scalafixConfig.in(config).value)
         val (shell, mainInterface0) = scalafixArgsFromShell(
           shellArgs,
           scalafixInterfaceProvider.value,
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -387,7 +387,7 @@ object ScalafixPlugin extends AutoPlugin {
       if (shellArgs.rules.isEmpty && shellArgs.extra == List("--help")) {
         scalafixHelp
       } else {
-        val scalafixConf = scalafixConfig.in(config).value
+        val scalafixConf = scalafixConfig.in(config).value.map(_.toPath)
         val (shell, mainInterface0) = scalafixArgsFromShell(
           shellArgs,
           scalafixInterfaceProvider.value,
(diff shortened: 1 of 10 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -89,6 +89,11 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional location to .scalafix.conf file to specify which scalafix rules should run. " +
           "Defaults to the build base directory if a .scalafix.conf file exists."
       )
+    val scalafixConfigOnCompile: SettingKey[Option[File]] =
```
bjaglin, line 92: reading that out loud, I am wondering if this wouldn't be better
```suggestion
    val scalafixOnCompileConfig: SettingKey[Option[File]] =
```
WDYT?

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -180,6 +185,7 @@ object ScalafixPlugin extends AutoPlugin {
 
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     scalafixConfig := None, // let scalafix-cli try to infer $CWD/.scalafix.conf
+    scalafixConfigOnCompile := scalafixConfig.value,
```
bjaglin, line 188: I am not sure this is doing what we want as _I think_ we won't capture the `scalafixConfig` values set at `ThisBuild` or project level. `scalafixConfigOnCompile` could remain undefined, and the fallback happens at evaluation time, like https://github.com/scalacenter/sbt-scalafix/pull/140/files#diff-b6db8e3499369917214568aee2dd0960R357. In that case, you wouldn't need the `Option` wrapper on the setting key type.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: I agree 👍

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: Thank you for pointing out that scalafixConfig in ThisBuild or project will not be captured (my bad).

And setting it as undefined and using `or` operator sounds better. I'll update it with tests.

taisukeoe: Since `or` operator doesn't work due to type parameter compatibility (SettingKey[File] v.s. SettingKey[Option[File]]) between scalafixOnCompileConfig and scalafixConfig, I'd use `?` operator for fallback implementation.

https://github.com/taisukeoe/sbt-scalafix/blob/92edd336cc87450ebc95d29e49684b8521ee89b2/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L355-L359

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
```diff
@@ -0,0 +1,37 @@
...
+-> example/scalafix
+-> example/scalafix --check RemoveUnused
+> example/compile
+## A statement incl. null has been removed by the precedence scalafix-on-compile.
+> example/scalafix
+
+# check if scalafix-on-compile uses .scalafix-on-compile.conf
+#   and scalafix-explicit-run uses .scalafix-explicit-run.conf
+
+> set scalafixConfig.in(ThisBuild) := Some(file(".scalafix-explicit-run.conf")),
```
bjaglin, line 25: just for the sake of exercising a bit more the scopes, we could do
```suggestion
> set scalafixConfig.in(example) := Some(file(".scalafix-explicit-run.conf")),
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -344,7 +349,14 @@ object ScalafixPlugin extends AutoPlugin {
...
       } else {
-        val scalafixConf = scalafixConfig.in(config).value.map(_.toPath)
+        val scalafixConf =
+          if (scalafixRunExplicitly.value) scalafixConfig.in(config).value
+          else
+            scalafixOnCompileConfig
+              .in(config)
+              .?
+              .value
+              .orElse(scalafixConfig.in(config).value)
```
bjaglin, line 359: I think you can also do
```suggestion
            scalafixConfig
              .in(config)
              .or(scalafixOnCompileConfig.in(config)(Some.apply))
              .value
```
Not sure if it's more readable though.

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
```diff
@@ -0,0 +1,37 @@
...
+
+# check if scalafix-on-compile uses .scalafix-on-compile.conf
+#   and scalafix-explicit-run still uses .scalafix.conf
+
+> set scalafixOnCompileConfig.in(ThisBuild) := file(".scalafix-on-compile.conf"),
+-> example/scalafix
+-> example/scalafix --check RemoveUnused
+> example/compile
+## A statement incl. null has been removed by the precedence scalafix-on-compile.
+> example/scalafix
```
bjaglin, line 20: to actually assert on the rewrite, you can use `newer`, see https://github.com/scalacenter/sbt-scalafix/blob/11d4c3f1bb182c2eb8c7b81ebcb6535f85a3f98e/src/sbt-test/skip-windows/caching/test#L42-L45

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
```diff
@@ -0,0 +1,37 @@
...
+> example/compile
+
+# check if scalafix-on-compile uses .scalafix-on-compile.conf
+#   but scalafix-explicit-run uses .scalafix.conf because scalafixConfig is None
+
+-> noneScalafixConfig/scalafix
+-> noneScalafixConfig/scalafix --check RemoveUnused
+> noneScalafixConfig/compile
+## A statement incl. null has been removed by the precedence scalafix-on-compile.
+> noneScalafixConfig/scalafix
```
bjaglin, line 36: same comment as above, an assertion on what changed would make it more readable

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
```diff
@@ -0,0 +1,37 @@
...
+
+> set scalafixOnCompileConfig.in(ThisBuild) := file(".scalafix-on-compile.conf"),
+-> example/scalafix
+-> example/scalafix --check RemoveUnused
+> example/compile
+## A statement incl. null has been removed by the precedence scalafix-on-compile.
+> example/scalafix
+
+# check if scalafix-on-compile uses .scalafix-on-compile.conf
+#   and scalafix-explicit-run uses .scalafix-explicit-run.conf
```
bjaglin, line 23: can you add an assertion that verifies that? wouldn't the test still pass if line 25 was removed?

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -89,6 +89,11 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional location to .scalafix.conf file to specify which scalafix rules should run. " +
           "Defaults to the build base directory if a .scalafix.conf file exists."
       )
+    val scalafixOnCompileConfig: SettingKey[File] =
+      settingKey[File](
+        "Optional location to .scalafix.conf file to specify which scalafix rules should run on compile. " +
```
bjaglin, line 94:
```suggestion
        "Location to .scalafix.conf file to specify which scalafix rules should run on compile. " +
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -89,6 +89,11 @@ object ScalafixPlugin extends AutoPlugin {
...
+          "Defaults to the same as scalafixConfig."
```
bjaglin, line 95:
```suggestion
          "scalafixOnCompile must be set for this to have an effect. Defaults to the same as scalafixConfig."
```

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
```diff
@@ -0,0 +1,37 @@
...
+# check if scalafix-on-compile is enabled and uses .scalafix.conf
+
+> set scalafixOnCompile.in(ThisBuild) := true
+-> example/compile
+-> example/scalafix
+
+# check if scalafix-on-compile uses .scalafix-on-compile.conf
+#   and scalafix-explicit-run still uses .scalafix.conf
+
+> set scalafixOnCompileConfig.in(ThisBuild) := file(".scalafix-on-compile.conf"),
```
bjaglin, line 15: this fails on sbt 0.13.x with the trailing comma
```suggestion
> set scalafixOnCompileConfig.in(ThisBuild) := file(".scalafix-on-compile.conf")
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -344,7 +349,14 @@ object ScalafixPlugin extends AutoPlugin {
       if (shellArgs.rules.isEmpty && shellArgs.extra == List("--help")) {
         scalafixHelp
       } else {
-        val scalafixConf = scalafixConfig.in(config).value.map(_.toPath)
+        val scalafixConf =
+          if (scalafixRunExplicitly.value) scalafixConfig.in(config).value
```
bjaglin, line 353: you can't totally rely on `scalafixRunExplicitly` as it won't be true when scalafix is potentially invoked through another plugin/task, so you should strengthen the check here
```suggestion
          if (scalafixRunExplicitly.value && !scalafixOnCompile.value) scalafixConfig.in(config).value
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: (I think I mixed up the order but you get the point)

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
taisukeoe: It was typo. Thank you!

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
taisukeoe: In this scripted test, `> example/scalafix` always runs only DisableSyntax, so that it won't rewrite anything.

Only `example/compile` or `nonScalafixConfig/compile` may rewrite. So I'll add some assertion about their rewrites.

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
taisukeoe: > set scalafixConfig.in(ThisBuild) := Some(file(".scalafix-explicit-run.conf")),

line 25(above) is mandatory for line 26, since `.scalafix-explicit-run.conf` and `.scalafix.conf` have different DisableSyntax config from each other.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: I'm afraid that `&& !scalafixOnCompile.value` won't work, because scalafixConfig is intended to be used in case scalafix is explicitly invoked, even if `scalafixOnCompile := true`. 

Please let me describe a use case:

Given that build and scalafix config files are as follows,

```
//build.sbt

scalafixOnCompile := true
scalafixOnCompileConfig := file(".scalafix-on-compile.conf")

semanticdbEnabled := true
semanticdbVersion := scalafixSemanticdb.revision

//.. and other settings
```
```
//.scalafix.conf
rules = [DisableSyntax]
DisableSyntax.keywords = [
  var
  null
]
```

```
//.scalafix-on-compile.conf
rules = [RemoveUnused]
```

On every-compile, a user run DisableSyntax rule for prompt lint without touching source codes.  Occasionally, the user run RemoveUnused rule to remove unused imports or variables by explicit scalafix run.

bjaglin: Apologies, that would rather be:
```suggestion
          if (scalafixRunExplicitly.value || !scalafixOnCompile.value) scalafixConfig.in(config).value
```

In your use-case, that would prevent `.scalafix-on-compile.conf` to be used when `scalafix` is called indirectly via a `lintAll` task defined in the build or in a plugin (in which case `scalafixRunExplicitly.value == false`).

bjaglin: (I am guessing you flipped `.scalafix.conf` and `.scalafix-on-compile.conf` in your example since `RemoveUnused` is more aggresive than `DisableSyntax` so it probably requires an explicit run - I didn't take that into account in my answer above)

bjaglin: > Apologies, that would rather be:
> 
> In your use-case, that would prevent `.scalafix-on-compile.conf` to be used when `scalafix` is called indirectly via a `lintAll` task defined in the build or in a plugin (in which case `scalafixRunExplicitly.value == false`).

Scratch the suggestion, actually it wouldn't prevent that. But there is still this corner case here that we can't really handle...

bjaglin: Maybe another way to implement the specific config file for compile, and for compile only, would be to pass `--config ${scalafixOnCompileConfig.in(config.value})` in https://github.com/scalacenter/sbt-scalafix/blob/11d4c3f1bb182c2eb8c7b81ebcb6535f85a3f98e/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L107. Since the CLI arguments are passed last https://github.com/scalacenter/sbt-scalafix/blob/11d4c3f1bb182c2eb8c7b81ebcb6535f85a3f98e/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L366, this might work!

taisukeoe: > when scalafix is potentially invoked through another plugin/task

Indeed. In current implementation, scalafix-invoked-by-another-plugin uses scalafixOnCompileConfig. Since scalafixOnCompileConfig is described as a conf file only for on-compilation hook, it may surprise users.

What do you think to define a task flag like this, and use in this PR?

```scala
 private val scalafixRunOnCompile: Def.Initialize[Task[Boolean]] =
    Def.task {
      executionRoots.value.exists (_.key == compile.key)
    }
```

bjaglin: > What do you think to define a task flag like this, and use in this PR?
> 
> ```scala
>  private val scalafixRunOnCompile: Def.Initialize[Task[Boolean]] =
>     Def.task {
>       executionRoots.value.exists (_.key == compile.key)
>     }
> ```

I had thought about it in the implementation of `scalafixOnCompile` - unfortunately you have the same problem: most of the times `compile` is invoked indirectly (via `test` for example) so this is not reliable.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: Thank you for following up!
It's good to know the way you showed, and I would prefer yours since I feel the `.?` line a bit ugly 😅

taisukeoe: Hmm, `scalafixOnCompileConfig.in(config).apply(Option.apply)` returns `Initalize[Option[File]]` which doesn't have `or` operator.
So it seems to require casting to `SettingKey[Option[File]]`...🤔 

```scala
scalafixOnCompileConfig
   .in(config)(Option.apply)
   .or(scalafixConfig.in(config)).value
```

```
value or is not a member of sbt.Def.Initialize[Option[sbt.File]]
[error] possible cause: maybe a semicolon is missing before `value or'?
[error]               .or(scalafixConfig.in(config)).value
[error]                ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

Anyway, `or` operator implementation uses `?` so that it is no big difference in semantics. So please let me go with current `.?` way.

bjaglin: Sure, apologies for the non-working proposal. Note that such a fallback won't be needed if we end up validating the other implementation https://github.com/scalacenter/sbt-scalafix/pull/148#discussion_r453842508, where `?` will be needed to conditionally pass an explicit config.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: Thank you for your throughout support on this.

> Maybe another way to implement the specific config file for compile, and for compile only, would be to pass --config ${scalafixOnCompileConfig.in(config.value})

I'd agree that this way would be the most reliable, so that I've rewrote on this way.

taisukeoe:
@bjaglin 
Thank you for your throughout review! I've reflected review comments.

NOTE:
Since I also find there is no way to do `--check` against `scalafixOnCompileConfig` 😱 , it might be a problem especially in CI/CD context.

So I've added `scalafixOnCompileCheck` InputKey, which would run `scalafix --check` against scalafixOnCompileConfig.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: @bjaglin No problem at all. Your suggestion helps my understanding a lot.

I've also confirmed https://github.com/scalacenter/sbt-scalafix/pull/148#discussion_r453842508 is working well and no need for `?` or `or` operators on this 👍

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
```diff
@@ -0,0 +1,52 @@
...
+
+# check if scalafix-on-compile uses .scalafix-on-compile.conf
+#   and scalafix-explicit-run still uses .scalafix.conf
+
+> set scalafixOnCompileConfig.in(ThisBuild) := file(".scalafix-on-compile.conf")
+-> example/scalafix --check
+-> example/scalafixOnCompileCheck
+$ copy-file example/src/main/scala/Example.scala noneScalafixConfig/src/test/scala/CopiedFile.scala
+> example/compile
+$ newer example/src/main/scala/Example.scala example/src/test/scala/CopiedFile.scala
```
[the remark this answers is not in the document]
bjaglin: I had suggested `newer` as a way to assert that `compile` did run Scalafix with the right Scalafix config, but it brings a lot of complexity. Maybe an explicit `scalafix --check RulesThatShouldHaveBeenApplied --settings.ThatShouldHaveBeenApplied` would be easier?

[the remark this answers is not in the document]
bjaglin: Where does `example/src/test/scala/CopiedFile.scala` comes from?

taisukeoe:
> Would it be possible to add a scripted to validate that this new implementation actually address #148 (comment) ?

Sounds good :+1:

> However, I am rather skeptical of scalafixOnCompileCheck:

Agree with your point about appropriate name and separating PullRequest.
Regarding to other two points, let's follow up on another PR.

src/sbt-test/sbt-scalafix/scalafixOnCompileConfig/test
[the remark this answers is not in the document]
taisukeoe: > Where does example/src/test/scala/CopiedFile.scala comes from? Would this assertion fail if the line above was removed?

Yes. I agree it's better not to depend on the previous state. So let's replace it with explicit `scalafix --check RULE`.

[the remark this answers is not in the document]
bjaglin: I was confused as the `copy-file` targeted `noneScalafixConfig`

bjaglin:
> Instead of implementing this here, I wonder if we can instead support a new `rulesOnCompile` field in `.scalafix.conf` 🤔 The `scalafixOnCompile` setting would then instead do something like `ScalafixArguments.withOnCompileEnabled(true)` and it would be the responsibility of `scalafix-cli` too use `rulesOnCompile` instead of `rules`.

Sounds good. I guess this can be done in a second step though as it wouldn't change the API here?

bjaglin:
> Instead of implementing this here, I wonder if we can instead support a new `rulesOnCompile` field in `.scalafix.conf`

Good idea. I started looking at integrating the version in .scalafix.conf, and I was wondering how we would handle 2 potentially different versions. This would solve this challenge. When possible, features should really be exposed via the configuration file rather than new sbt knobs.

bjaglin:
@taisukeoe Unfortunately that generic approach makes a lot of your work on this PR obsolete... Would you be interested in implementing the feature in scalafix itself as suggested? I can provide help on PRs or gitter.

taisukeoe:
@olafurpg
Thank you for sharing your insight.

>  it would be the responsibility of scalafix-cli too use rulesOnCompile instead of rules.

Now I agree with making `scalafix-cli` manage on-compile stuff.
Since `scalafix-cli` already manages some compilation stuff, it should be fine in terms of keeping `scalafix-cli` responsibility reasonable.

However, I'm on the fence to introduce `rulesOnCompile`, because settings in each rule can also be different between on-compile and non-compile.

For example, there are several rules like [MissingFinal in vovapolu/scaluzzi](https://github.com/vovapolu/scaluzzi/blob/387d7f41fb5288bfcdcb29a11e7363edc739fe1f/scalafix/rules/src/main/scala/scalafix/internal/scaluzzi/MissingFinal.scala#L34) which have their own configs about diagnostics and rewrites. 
It seems to be a good use-case that users enable only their diagnostics on compile and enable their rewrites in non-compile, while this usecase cannot be done by `rulesOnCompile`.

Rather, introducing `on-compile` block and put all fields (rules and settings in each rule) about on-compile into the block might make sense IMHO.

Based on the assumption, I've not thought through enough which is better - putting all together in a scalafix.conf, or splitting into .scalafix.conf and .scalafix-on-compile.conf.

taisukeoe:
@bjaglin

> @taisukeoe Unfortunately that generic approach makes a lot of your work on this PR obsolete... Would you be interested in implementing the feature in scalafix itself as suggested? I can provide help on PRs or gitter.

I'm completely fine with going the better option :+1:
I believe the discussion on this PullRequest also helps us reach the better goal.

I'd still be interested in contributing scalafix-cli stuff on this. So I'll setup another PullRequest separately.

taisukeoe:
Since I've prepared a PullRequest https://github.com/scalacenter/scalafix/pull/1217 , please let me close this PR.

PR #152, opened 2020-07-15
fix usage of scalafix via another task

bjaglin:
Regression of 0cf5501: initial (cold) invocations to a task delegating to `scalafix` would result in `scalafix` being called before compilation (because `scalafixRunExplicitly == false`) causing false negatives.

This makes sure compilation is always triggered before `scalafix` (no matter how it's called), but depends on the final sub-tak of `compile` in a cycle-free way.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -100,9 +100,7 @@ object ScalafixPlugin extends AutoPlugin {
         compile := Def.taskDyn {
           val oldCompile =
             compile.value // evaluated first, before the potential scalafix evaluation
-          val runScalafixAfterCompile =
-            scalafixOnCompile.value && !scalafixRunExplicitly.value
-          if (runScalafixAfterCompile)
+          if (scalafixOnCompile.value)
             scalafix
               .toTask("")
               .map(_ => oldCompile)
@@ -415,7 +413,7 @@ object ScalafixPlugin extends AutoPlugin {
       ).findErrors(files, dependencies, withScalaInterface)
       if (errors.isEmpty) {
         val task = Def.task {
-          // passively consume compilation output without triggering compile as it can result in a cyclic dependency
+          // don't use fullClasspath as it results in a cyclic dependency via compile when scalafixOnCompile := true
           val classpath =
             dependencyClasspath.in(config).value.map(_.data.toPath) :+
               classDirectory.in(config).value.toPath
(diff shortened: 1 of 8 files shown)


src/sbt-test/sbt-scalafix/wrapper/test
```diff
@@ -0,0 +1,14 @@
+# check that missing rewrites are detected, with or without a warm compilation cache
+> set scalafixOnCompile := false
+-> lintAll
```
bjaglin, line 3: on master
```
[info] [info] Non-compiled module 'compiler-bridge_2.12' for Scala 2.12.12. Compiling...
[info] [info] Running scalafix on 2 Scala sources
[error] /tmp/sbt_4021633e/wrapper/src/main/scala/example/Example.scala:3: error: not found: object imported
[error] import imported.Imported
[error]        ^
[info] [info]   Compilation completed in 14.904s.
[info] [warn] /tmp/sbt_4021633e/wrapper/src/main/scala/example/Example.scala:3:17: Unused import
[info] [warn] import imported.Imported
[info] [warn]                 ^
[info] [warn] one warning found
[info] [info] Done compiling.
[info] [success] Total time: 20 s, completed Jul 15, 2020, 11:52:59 AM
[error] x sbt-scalafix / wrapper 
[error]  Cause of test exception: {line 3}  Command succeeded but failure was expected
```

bjaglin: I investigated where the concurrent `Non-compiled module 'compiler-bridge_2.12' for Scala 2.12.12. Compiling...`/`Done compiling` was coming from as it looked like it could race with the bytecode consumption in `scalafix`. It turns out to be from `Test / scalafix`, so I added caab820 to strengthen the test.

Without the fix, the test logs now shows:
```
[info] [info] Running scalafix on 2 Scala sources
[error] /tmp/sbt_93ed664/wrapper/src/main/scala/example/Example.scala:3: error: not found: object imported
[error] import imported.Imported
[error]        ^
[info] [success] Total time: 5 s, completed Jul 15, 2020, 5:29:15 PM
[error] x sbt-scalafix / wrapper 
[error]  Cause of test exception: {line 3}  Command succeeded but failure was expected

```

bjaglin:
(looks like GitHub is having another outage, and my second commit is not visible here)

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -428,8 +426,8 @@ object ScalafixPlugin extends AutoPlugin {
             streams.in(config, scalafix).value
           )
         }
-        if (scalafixRunExplicitly.value) task.dependsOn(compile.in(config))
-        else task
+        // sub-task of compile after which bytecode should not be modified
+        task.dependsOn(manipulateBytecode.in(config))
```
olafurpg, line 430: 👍

PR #153, opened 2020-07-16
prevent concurrent modifications of sources by limiting concurrency

bjaglin:
When rewrite rules are executed, `scalafix` and `scalafixAll` may conflict with other tasks transforming `unmanagedSources`. Since there is no way to identify such tasks given the sbt model, and to be defensive against this scenario, `scalafix` tasks may only run concurrently with other `scalafix` tasks, but not with any other tasks.

For explicit invocations, this should not damage throughput, but for complex builds with `scalafixOnCompile := true`, this can have a limited impact during the first compile; after that caching should make it negligible.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -23,7 +23,6 @@ object ScalafixPlugin extends AutoPlugin {
   override def requires: Plugins = JvmPlugin
 
   object autoImport {
-    @deprecated("not used internally, use your own tag", "0.9.17")
     val Scalafix = Tags.Tag("scalafix")
 
     val ScalafixConfig = config("scalafix")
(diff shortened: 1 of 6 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -375,6 +378,9 @@ object ScalafixPlugin extends AutoPlugin {
         }
       }
     }
+    task.tag(scalafixTag)
```
bjaglin, line 381:
```suggestion
    task
```
breaks the scripted
```
[info] [error] (Compile / check) java.lang.AssertionError: assertion failed: missing footer in
[info] [error] <<<// HEADER
[info] [error] object Example
[info] [error] >>>
```

bjaglin:
@sjrd I am proposing this to address the concerns you expressed in https://github.com/scalacenter/sbt-scalafix/pull/140#issuecomment-656663541
> modifying files in a task (either input files or files produced by earlier tasks) is a very bad thing, and you should never do that, ever. The parallelism of sbt means that things can execute concurrently on already created files, causing very bad issues. If you want to modify things, use a command instead (and no, you can't execute a command from a task, not in a sound way anyway).

bjaglin:
> FYI I’ll be away for a few weeks and may not comment much on GitHub

Enjoy your break! I'll also be AFK some weeks over the next 5 ones.

PR #154, opened 2020-07-25
Respect Configuration in scalafixConfigSettings

etspaceman:
Hello - 

Upgrading to sbt-scalafix 0.9.19 caused one of my projects to fail. When poking around, I noticed that the new `scalafixOnCompile` configuration setup did not respect the configuration that was passed into the `scalafixConfigSettings` function. 

The scripted test got around this by doing `inConfig(IntegrationTest)(scalafixConfigSettings(IntegrationTest))`. You should be able to just do ` scalafixConfigSettings(IntegrationTest)`. 

This PR updates the `scalafixConfigSettings` function to respect the passed `Configuration` value for these items.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -96,12 +96,12 @@ object ScalafixPlugin extends AutoPlugin {
 
     def scalafixConfigSettings(config: Configuration): Seq[Def.Setting[_]] =
       Seq(
-        scalafix := scalafixInputTask(config).evaluated,
+        (scalafix in config) := scalafixInputTask(config).evaluated,
         (compile in config) := Def.taskDyn {
           val oldCompile =
             (compile in config).value // evaluated first, before the potential scalafix evaluation
-          if ((scalafixOnCompile in config).value)
-            scalafix
+          if (scalafixOnCompile.value)
+            (scalafix in config)
               .toTask("")
               .map(_ => oldCompile)
           else Def.task(oldCompile)
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -95,25 +95,26 @@ object ScalafixPlugin extends AutoPlugin {
       "org.scalameta" % "semanticdb-scalac" % scalametaVersion cross CrossVersion.full
 
     def scalafixConfigSettings(config: Configuration): Seq[Def.Setting[_]] =
-      Seq(
-        (scalafix in config) := scalafixInputTask(config).evaluated,
-        (compile in config) := Def.taskDyn {
-          val oldCompile =
-            (compile in config).value // evaluated first, before the potential scalafix evaluation
-          if (scalafixOnCompile.value)
-            (scalafix in config)
-              .toTask("")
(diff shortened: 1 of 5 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -96,12 +96,12 @@ object ScalafixPlugin extends AutoPlugin {
...
-        scalafix := scalafixInputTask(config).evaluated,
-        compile := Def.taskDyn {
+        (scalafix in config) := scalafixInputTask(config).evaluated,
+        (compile in config) := Def.taskDyn {
           val oldCompile =
-            compile.value // evaluated first, before the potential scalafix evaluation
+            (compile in config).value // evaluated first, before the potential scalafix evaluation
           if (scalafixOnCompile.value)
-            scalafix
+            (scalafix in config)
```
bjaglin, line 104: we could wrap the `Seq` inside `inConfig(config)` instead of repeating the scope for each key

etspaceman:
Hey @bjaglin,

> Which version were you upgrading from, and what started to fail? 

0.9.17. I basically had the following code snippet:

```
.settings(scalafixConfigSettings(IntegrationTest):_*)
```

This actually worked for a long time. Scalafix ran properly on my project's integration tests. I must have missed that item in the docs when going through my setup originally, and I was happily oblivious. 😅 

>  I am wondering what can have regressed in 0.9.19?

Upgrading to 0.9.19 introduced the `scalafixOnCompile` option which added the `compile` section of the settings above. The `compile` step is undefined. You can reproduce this in the above PR's tests by reverting the plugin's change:

```
[info] [error] Reference to undefined setting:
[info] [error]
[info] [error]   rewrite / compile from rewrite / compile ((scalafix.sbt.ScalafixPlugin.autoImport.scalafixConfigSettings) ScalafixPlugin.scala:100)
[info] [error]      Did you mean rewrite / Compile / compile ?
```

> I am AFK at the moment, but I think it is worth investigating whether we could remove that argument

Potentially! That is indeed the original source of my confusion here, because I assumed that passing in `IntegrationTest` would properly scope that function call. Having it in the signature makes this quite confusing otherwise.

That said, I would also be happy with this implementation. I am biased of course. 😄

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
etspaceman: addressed

src/sbt-test/sbt-scalafix/cross-build/build.sbt
```diff
@@ -41,6 +40,8 @@ lazy val scala212 = project
     scalaVersion := Versions.scala212,
     scalafixSettings
   )
+  .settings(scalafixConfigSettings(IntegrationTest): _*)
```
bjaglin, line 43: `inConfig` seems to return the same type as `scalafixConfigSettings`, is there a reason why you moved down all the settings and used the spread operator? it would be nicer if we could keep them inline (for the diff, but mostly for users!)

bjaglin:
Sorry for the silence, summer break here. Since the change is backward-compatible and I have found at least one project using this syntax ([akka-grpc](https://doc.akka.io/docs/akka-grpc/current/buildtools/sbt.html#configurations)), I am fine with it - I just have a remaining cosmetic comment.

@olafurpg any objection?

PR #159, opened 2020-08-25
Change scalafixResolvers to come from Coursier defaults so can be overridden

scottrice10:
The impetus for this proposed change is trying to use privately hosted `scalafix` rules with [`scala-steward`](https://github.com/scala-steward-org/scala-steward). Currently, the scalafix tutorial describes how to [add custom scalafix resolvers](https://scalacenter.github.io/scalafix/docs/developers/tutorial.html#adding-custom-resolver) in a project's `build.sbt`. Since scala-steward adds the sbt-scalafix plugin on the fly, however, it would further need to edit a project's `build.sbt` in order to add custom resolvers, which would be tricky.

The alternative I'm proposing here sets the `scalafixResolvers` to Coursier's [`defaultRepositories`](https://github.com/coursier/coursier/blob/master/modules/coursier/jvm/src/main/scala/coursier/PlatformResolve.scala#L19).  The [defaults](https://github.com/coursier/coursier/blob/master/modules/coursier/jvm/src/main/scala/coursier/PlatformResolve.scala#L60-L63) are the same as previously set - `Repository.ivy2Local()` and `Repository.central()`. But using `defaultRepositories` allows overriding the repositories via the `COURSIER_REPOSITORIES` environment variable, as documented [here](https://get-coursier.io/docs/other-repositories) and shown in the `defaultRepositories` code.

I've tested this code locally with first doing 
```
export COURSIER_REPOSITORIES="ivy2Local|central|ivy:https://a.b.c/d/f/[organisation]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]"
```
and then running `sbt-scalafix` on a repo (without overriding the `scalafixResolvers`), and verified it picks up the `scalafix` rules which exist only in the private ivy artifactory resolver.

When trying to run `scala-steward` to apply the privately hosted `scalafix` rules, the current error I'm getting is:
```
13:21:47  [error] (root/*:scalafixAll) scalafix.sbt.InvalidArgument: Failed to fetch [Dependency(Module(a.b.c, scalafix_2.12), 0.0.1-12-f9db091-SNAPSHOT)]from [IvyRepository(file:/home/user/.ivy2/local/[organisation]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext], dropInfoAttributes = true), MavenRepository(https://repo1.maven.org/maven2), MavenRepository(https://oss.sonatype.org/content/repositories/public), MavenRepository(https://oss.sonatype.org/content/repositories/snapshots)]
```
which I expect this proposed change to resolve. If this change is accepted, I will update the [`scala-steward` running docs](https://github.com/scala-steward-org/scala-steward/blob/master/docs/running.md#private-repositories) to explain how adding `export COURSIER_REPOSITORIES="ivy2Local|central|somePrivateRepository"` before running `scala-steward` allows applying privately hosted `scala-fix` rules.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -180,13 +180,21 @@ object ScalafixPlugin extends AutoPlugin {
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     scalafixConfig := None, // let scalafix-cli try to infer $CWD/.scalafix.conf
     scalafixOnCompile := false,
-    scalafixResolvers := Repository.defaults().asScala.toSeq ++ Seq(
-      coursierapi.MavenRepository
-        .of("https://oss.sonatype.org/content/repositories/public"),
-      coursierapi.MavenRepository.of(
-        "https://oss.sonatype.org/content/repositories/snapshots"
-      )
-    ),
+    scalafixResolvers :=
+      // Repository.defaults() defaults to Repository.ivy2Local() and Repository.central(). These can be overridden per
+      // env variable, e.g., export COURSIER_REPOSITORIES="ivy2Local|central|sonatype:releases|jitpack|https://corporate.com/repo".
+      // See https://github.com/coursier/coursier/blob/master/modules/coursier/jvm/src/main/scala/coursier/PlatformResolve.scala#L19-L68
+      // and https://get-coursier.io/docs/other-repositories for more details.
+      // Also see src/sbt-test/sbt-scalafix/scalafixResolvers/test for a scripted test preserving this behavior.
+      Repository.defaults().asScala.toSeq ++
+        Seq(
+          coursierapi.MavenRepository.of(
+            "https://oss.sonatype.org/content/repositories/public"
+          ),
+          coursierapi.MavenRepository.of(
+            "https://oss.sonatype.org/content/repositories/snapshots"
+          )
+        ),
     scalafixDependencies := Nil,
     commands += ScalafixEnable.command,
     scalafixInterfaceProvider := ScalafixInterface.fromToolClasspath(
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -15,6 +15,7 @@ import scalafix.interfaces.ScalafixError
 import scalafix.internal.sbt.Arg.ToolClasspath
 import scalafix.internal.sbt._
 
+import scala.collection.JavaConverters.collectionAsScalaIterableConverter
 import scala.util.Try
 import scala.util.control.NoStackTrace
 
@@ -179,9 +180,7 @@ object ScalafixPlugin extends AutoPlugin {
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     scalafixConfig := None, // let scalafix-cli try to infer $CWD/.scalafix.conf
     scalafixOnCompile := false,
-    scalafixResolvers := Seq(
-      Repository.ivy2Local(),
-      Repository.central(),
+    scalafixResolvers := Repository.defaults().asScala.toSeq ++ Seq(
       coursierapi.MavenRepository
         .of("https://oss.sonatype.org/content/repositories/public"),
       coursierapi.MavenRepository.of(
--- a/src/sbt-test/sbt-scalafix/scalafixResolvers/test
+++ b/src/sbt-test/sbt-scalafix/scalafixResolvers/test
@@ -1,0 +1,3 @@
+# reload to ensure env variable COURSIER_REPOSITORIES is picked up by scalafixResolvers
+> reload
+> check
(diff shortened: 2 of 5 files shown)


scottrice10:
I'll add, this would be for running `scalafix` rules using the [`dependency` schema](https://scalacenter.github.io/scalafix/docs/rules/external-rules.html), i.e.:
```
scalafix dependency:someRule@com.org:scalafix:0.0.1-12-f9db091-SNAPSHOT
```

I also tried using the [`http` / `github` schemas](https://scalacenter.github.io/scalafix/docs/developers/sharing-rules.html#using-http) to apply privately hosted scalafix rules, and ran into a different blocker.  You can use the http method with privately hosted code following these steps:
Steps:
```
1. While logged into Github, in the browser go the one of the rule files, such as https://github.com/org/repo/blob/master/scalafix/rules/Imports.scala, click the "Raw" button.
1. Copy the browser url for the raw option, including the ?token=<sessionToken> at the end.
1. Run:
scalafix --rules=https://raw.githubusercontent.com/org/repo/master/rules/Imports.scala?token=<sessionToken>
...

That token is a session token, though, not a personal access token, so it expires after a few days. You can use a Github personal access token to get raw github files using the api and passing headers, like:
```
  curl -H 'Authorization: token <personalAccessToken>' \
  -H 'Accept: application/vnd.github.v4.raw' \
  -L https://api.github.com/repos/org/repo/contents/rules/Imports.scala
```
To make this work, `scalafix` would need to get updated to accept headers used to fetch the raw github file, and `scala-steward`'s configuration would need to get updated to pass an access token in the headers.

**Comment by scottrice10** [comment]
@bjaglin @mlachkar @olafurpg WDYT?

**Review by olafurpg — APPROVED** [review]
Thank you for this contribution! LGTM 👍 Any objections @mlachkar  @bjaglin ?

**Review by bjaglin — APPROVED** [review]
LGTM, thanks for the detailed explanation!

What would be great would be to document & prevent regressions against that in a scripted test that would run a rule available via a non-standard, public resolver declared in `COURSIER_REPOSITORIES`. I don't know of any though, do you by any chance?

File: src/main/scala/scalafix/sbt/ScalafixPlugin.scala [hunk 1]
```diff
@@ -179,9 +180,7 @@ object ScalafixPlugin extends AutoPlugin {
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     scalafixConfig := None, // let scalafix-cli try to infer $CWD/.scalafix.conf
     scalafixOnCompile := false,
-    scalafixResolvers := Seq(
-      Repository.ivy2Local(),
-      Repository.central(),
+    scalafixResolvers := Repository.defaults().asScala.toSeq ++ Seq(
```
arkban, line 183: Should this only happen if `useCoursier := false` (from the [SBT docs](https://www.scala-sbt.org/1.x/docs/sbt-1.3-Release-Notes.html#Library+management+with+Coursier)).

Maybe it makes more sense to just to tap into [SBT's default resolvers](https://www.scala-sbt.org/1.x/docs/Resolvers.html)?

scottrice10:
> ...via a non-standard, public resolver declared in COURSIER_REPOSITORIES. I don't know of any though, do you by any chance?

I agree this would be a good test. But no, I'm aware of a public non-standard resolver that hosts public scalafix rules. I took a look through the [public scalafix rules](https://github.com/scala-steward-org/scala-steward/blob/master/modules/core/src/main/resources/scalafix-migrations.conf) scala-steward is adding by default, but didn't find any that are publishing to resolver other than sonatype.

I setup the test to first remove sonatype, and then re-add it back via the env variable override, which I think demonstrates it working as intended.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
scottrice10: > Should this only happen if useCoursier := false (from the SBT docs).

As of now, as you mention, `scalafix` is bypassing the `sbt` resolvers, so the sbt setting `useCoursier := false` doesn't apply to setting the `scalafixResolvers`. This may come into play if your next suggestion is implemented, though?

> Maybe it makes more sense to just to tap into SBT's default resolvers?

For the particular problem I'm trying to solve at Rally, yes, this would work without needing an env variable override. Assuming all scala projects are using the `rally-sbt-plugin` (private to my company), which puts https://<customRepositoryHost>/ivy-libs-snapshot in `globalSettings`, and our private scalafix rules are published to the same https://<customRepositoryHost>/ivy-libs-snapshot, then using SBT's default resolvers for each repo would work for a tool like `scala-steward`.

Since `scala-steward` adds `sbt-scalafix` to repos on the fly, I could see a scenario where the repos' configured SBT resolvers are not the same as the custom resolver where the scalafix rules are hosted, in which case the env variable would still be desired to add additional custom resolvers.

scottrice10:
> I setup the test to first remove sonatype, and then re-add it back via the env variable override.

Ahh, this doesn't work. I think it works locally because my ivy2 local is already populated with the dependencies needed from sonatype. So, think I do need to find a public non-standard resolver that hosts public scalafix rules in order to properly test this...

scottrice10:
@bjaglin Do you think this test is sufficient?

src/sbt-test/sbt-scalafix/scalafixResolvers/test
```diff
@@ -0,0 +1,3 @@
+# reload to ensure env variable COURSIER_REPOSITORIES is picked up by scalafixResolvers
+> reload
```
bjaglin, line 2: out of curiosity, is the reload required by `sbt-dotenv` for the env file to be applied?

bjaglin (approved):
Wonderful, thank you for figuring out a way to prevent this from regressing!

PR #168, opened 2020-10-05
sbt 1.4.0

bjaglin:
The second commit works around https://github.com/sbt/sbt/issues/5647, which can be observed when the client uses sbt 1.4.0 (independently of the version we build with/against here) and defines `scalafixCaching`.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -97,7 +97,11 @@ object ScalafixPlugin extends AutoPlugin {
     def scalafixConfigSettings(config: Configuration): Seq[Def.Setting[_]] =
       inConfig(config)(
         Seq(
-          scalafix := scalafixInputTask(config).evaluated,
+          scalafix := {
+            // force detection of usage of `scalafixCaching` to workaround https://github.com/sbt/sbt/issues/5647
+            val _ = scalafixCaching.?.value
+            scalafixInputTask(config).evaluated
+          },
           compile := Def.taskDyn {
             val oldCompile =
               compile.value // evaluated first, before the potential scalafix evaluation
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.3.13
+sbt.version=1.4.0


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -97,7 +97,11 @@ object ScalafixPlugin extends AutoPlugin {
     def scalafixConfigSettings(config: Configuration): Seq[Def.Setting[_]] =
       inConfig(config)(
         Seq(
-          scalafix := scalafixInputTask(config).evaluated,
+          scalafix := {
+            // force detection of usage of `scalafixCaching` to workaround https://github.com/sbt/sbt/issues/5647
+            val _ = scalafixCaching.?.value
```
bjaglin, line 102:
```diff
 sbt:sbt-scalafix> inspect actual scalafixCaching
 [info] Setting: Boolean = true
 [info] Description:
 [info] 	Cache scalafix invocations (off by default, on if scalafixOnCompile := true).
 [info] Provided by:
 [info] 	ProjectRef(uri("file:/home/bjaglin/git/projects/sbt-scalafix/"), "sbt-scalafix") / scalafixCaching
 [info] Defined at:
 [info] 	/home/bjaglin/git/projects/sbt-scalafix/build.sbt:48
+[info] Reverse dependencies:
+[info] 	Test / scalafix
+[info] 	Compile / scalafix
 [info] Delegates:
 [info] 	scalafixCaching
 [info] 	ThisBuild / scalafixCaching
 [info] 	Global / scalafixCaching
```

adpi2 (approved):
This seems legitimate.

I wonder if this workaround is still needed though:
https://github.com/scalacenter/sbt-scalafix/blob/5238ad749e06059e2bb7c200aec8c8c7375ce76e/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L324-L328

The task graph would be simplified with a single level of *dynamicity* instead of two.

bjaglin:
> I wonder if this workaround is still needed though:

Unfortunately, even if it was fixed in 1.4.0 (I don't think it is), we would still need to keep the hack for compatibility with older 1.x clients. The first commit bumps the version of sbt for this build, but the sbt API build against and the runtime we run scripted with remains unchanged, see: https://github.com/scalacenter/sbt-scalafix/blob/5238ad749e06059e2bb7c200aec8c8c7375ce76e/build.sbt#L53-L59.

PR #169, opened 2020-10-10
add --triggered support when scalafixOnCompile is enabled

taisukeoe:
This is a companion Pull Request of https://github.com/scalacenter/scalafix/pull/1217 .

This adds `--triggered` support when `scalafixOnCompile` is enabled.


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -50,6 +50,12 @@ object ScalafixPlugin extends AutoPlugin {
           "but can be disabled with `scalafixCaching := false`."
       )
 
+    val scalafixArgsOnCompile: TaskKey[Seq[String]] =
+      taskKey[Seq[String]](
+        "Pass extra arguments to scalafix only when scalafix is triggred by compilation.  " +
+          "Only --triggered argument is passed by default."
+      )
```
taisukeoe, line 57: `scalafixArgsOnCompile` TaskKey is intended to be more extensible than just passing `--triggered`.

For example:
 - `scalafixArgsOnCompile += "--check"` allows to lint on compilation and rewrite on explicit invocation in the same `.scalafix.conf`.
 - Preferring `TaskKey` over `SettingKey` because `TaskKey` can rely on other `TaskKey` values, while `scalafixArgsOnCompile` is rarely referred by other `SettingKey`s. (Analogous to `scalacOptions`)

taisukeoe:
Question: what is the best way to test `sbt-scalafix` branch which depends on another wip branch on `scalacenter/scalafix`?

Currently I change `scalafixVersion` in `project/Dependencies.scala` to scalafix SNAPSHOT version which is publish locally. This works in certain scripted tests (in this case `scripted sbt-scalafix/scalafixOnCompile`), but doesn't work in tests which depend on `BuildInfo.scalafixVersion` explicitly (e.g. `scripted sbt-scalafix/local-rules` ).

bjaglin:
> Question: what is the best way to test `sbt-scalafix` branch which depends on another wip branch on `scalacenter/scalafix`?
> 
> Currently I change `scalafixVersion` in `project/Dependencies.scala` to scalafix SNAPSHOT version which is publish locally. This works in certain scripted tests (in this case `scripted sbt-scalafix/scalafixOnCompile`), but doesn't work in tests which depend on `BuildInfo.scalafixVersion` explicitly (e.g. `scripted sbt-scalafix/local-rules` ).

There will be a SNAPSHOT published once https://github.com/scalacenter/scalafix/pull/1217 is merged, so you should be able to get the CI green at that point. For now, I guess you could hardcode the SNAPSHOT in https://github.com/scalacenter/sbt-scalafix/blob/8284c35c07418bcf07095097ba7729a360ecf891/src/sbt-test/sbt-scalafix/local-rules/build.sbt#L24 as well. We will then hold merging this PR until we bump the to-be-tagged scalafix version on master, at which point I will remove the SNAPSHOTs, merge this, and tag sbt-scalafix.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: See my comment in https://github.com/scalacenter/scalafix/pull/1217#pullrequestreview-506174142, I am questionning the need for such a "open" customization point if it's just for `--check`. Do you have something else in mind?

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -107,7 +113,7 @@ object ScalafixPlugin extends AutoPlugin {
               compile.value // evaluated first, before the potential scalafix evaluation
             if (scalafixOnCompile.value)
               scalafix
-                .toTask("")
+                .toTask(scalafixArgsOnCompile.value.mkString(" ", " ", ""))
```
bjaglin, line 116: it would be nice to update the scripted test about caching, to make sure that an explicit `scalafix` after a compile with `scalafixOnCompile` does run rules that are not defined in the `triggered` overlay.

Given the current cache implementation, I think an explicit `scalafix` invocation will no longer benefit from the cache of the previous `compile` invocations after this PR. That's safer, as we can't assume that the configuration is effectively the same and add `--triggered` in https://github.com/scalacenter/sbt-scalafix/blob/8284c35c07418bcf07095097ba7729a360ecf891/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L510-L511. I guess we could do so if we see that there is no `triggered` section in the configuration.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: I'd think several alternatives of arguments on compilation in my mind.

1. `--triggered --check` : This should work well in most use cases, but it might be confusing if `.scalafix.conf` gets bigger and a lot of difference in configuration between `triggered` and explicit invocation.

2. `--config .scalafix.oncompile.conf` : In case `.scalafix.conf` is already big, or overlaying settings will make huge difference between on-compile or not, it would be easier to divide-and-conquer scalafix.conf file rather than `overlaying`.

3. `--check` : This makes `scalafixOnCompile` less powerful, but more simple. I'd think this is just a matter of preference, so I wouldn't push this use case so much.

Disclaimer: these alternatives are based on my assumptions. While I haven't seen **HUGE** `.scalafix.conf` which has to be divided and conquered for now, it might be chicken or egg.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: Thank you for your suggestion. I'll check how cache works later!

bjaglin:
@taisukeoe I hope you are well! Any update on this? Are you reconsidering the need? (in which case we could deprecate/clean up the parts in the scalafix repo)

taisukeoe:
@bjaglin 
Hi, thank you for pinging me. I'll try to finish this PR in this weekend.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: I have reconsidered this customization point, and removed it for now.
There are few rules which may break source codes in running on-compile, and can be managed only with `--triggered`.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
taisukeoe: `--triggered` is considered as a cache key only if there is `triggered` section in `.scalafix.conf`.

taisukeoe:
@bjaglin I'm very sorry for making you wait. Now this PR is ready. Would you please review this when you have time?

src/sbt-test/skip-windows/caching/test
```diff
@@ -106,6 +106,40 @@ $ exec chmod 000 src/main/scala/InitiallyValid.scala
...
+> set scalafixConfig := Some(file("files/Triggered.scalafix.conf"))
+$ mkdir src/main/scala
+$ copy-file files/Valid.scala src/main/scala/Valid.scala
+> scalafix
+$ exec chmod 000 src/main/scala/Valid.scala
+> scalafix --check
+-> scalafix --triggered --check
+$ delete src/main/scala
+
+# files should not be re-checked in scalafix explicit invocation after scalafix runs onCompile if any triggered configuration is not defined
```
bjaglin, line 131:
```suggestion
# files should not be re-checked in scalafix explicit invocation after scalafix runs onCompile if no triggered configuration is defined
```

src/sbt-test/sbt-scalafix/scalafixOnCompile/build.sbt
```diff
@@ -21,3 +21,8 @@ lazy val rewrite = project
     addCompilerPlugin(scalafixSemanticdb)
   )
   .settings(scalafixConfigSettings(IntegrationTest): _*)
+
+lazy val args = project
```
bjaglin, line 25: what does args refer to? what about
```suggestion
lazy val triggered = project
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -529,13 +530,29 @@ object ScalafixPlugin extends AutoPlugin {
                 // custom tool classpaths might contain directories for which we would need to stamp all files, so
                 // just disable caching for now to keep it simple and to be safe
                 throw StampingImpossible
+              case "--triggered" =>
+                // If there is a triggered section in a default config file, --triggered flag should be accounted.
```
bjaglin, line 534: looking at the implementation, it seems to work also for non-default configurations?
```suggestion
                // If there is a triggered section in the config file, --triggered flag should be accounted.
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -45,7 +45,8 @@ object ScalafixPlugin extends AutoPlugin {
 
     val scalafixOnCompile: SettingKey[Boolean] =
       settingKey[Boolean](
-        "Run Scalafix rule(s) declared in .scalafix.conf on compilation and fail on lint errors. " +
+        "Run Scalafix rule(s) declared in .scalafix.conf on compilation and fail on lint errors," +
+          "with overlaying configuration by `triggered` sections if defined in .scalafix.conf file." +
```
bjaglin, line 49: The plural in "sections" confused me a bit. I was using "overlay' as a verb in our previous discussions but I wonder if "overriden" isn't more clear in that context. What about:
```suggestion
        "Run Scalafix rule(s) declared in .scalafix.conf on compilation and fail on lint errors. " +
          "Default rules & rule configuration can be overriden using the `triggered` section." +
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -529,13 +530,29 @@ object ScalafixPlugin extends AutoPlugin {
...
             write(cacheKeys)
           case Arg.NoCache =>
             throw StampingImpossible
         }
 
+        private lazy val checkIfTriggeredSectionExists: Boolean = {
+          val confInArgs = interface.args
+            .collect { case Arg.Config(conf) => conf }
+            .flatten
+            .lastOption
```
bjaglin, line 548: :+1:

src/sbt-test/skip-windows/caching/files/Triggered.scalafix.conf
```diff
@@ -0,0 +1,4 @@
+rules = [DisableSyntax]
+DisableSyntax.noVars = true
+
+  triggered.DisableSyntax.noNulls = true
```
bjaglin, line 4: :+1: might be worth adding a comment just above to highlight that the whitespaces are intentional (even though the commit already says it)

bjaglin (commented):
Thanks for taking this to the finish line! I really appreciate the extensive testing :bow:  Structure and code LGTM! Please see a few naming/cosmetic suggestions.

Next step would be to document this in the website (https://scalacenter.github.io/scalafix/docs/users/configuration.html looks like a good place?) - unfortunately that's on the scalafix repo!

taisukeoe:
@bjaglin Thank you for your quick review! I've modified wordings.
As for documentation, is it also worth noting here?
https://scalacenter.github.io/scalafix/docs/users/installation.html#run-scalafix-automatically-on-compile

I think I can finish documentation in this couple of days, by next weekend at the latest.

taisukeoe:
@bjaglin 
I've created a doc PR. Would you please review this? Thanks!
https://github.com/scalacenter/scalafix/pull/1375

bjaglin (approved):

PR #173, opened 2020-11-04
bump to v0.9.22

mlachkar:


bjaglin:
Looking at the failure...

bjaglin:
@mlachkar I have pushed a fix, but I wonder if it's really worth introducing a breaking change in upstream scalafix testkit by having that new mandatory property https://github.com/scalacenter/scalafix/commit/d31b5dca1fdfd302256238af658f4016006f40c6#diff-37c85fdbae8ace086d947ad0a7c137d5683791485b006a4e15ac8a38fa0db375R99, as it's only concatenated with the value of an existing one. We could do just like https://github.com/scalacenter/sbt-scalafix/commit/cd9efb1b1e16f178da6be72a4cc621524e8a301a upstream (which I guess I should have copied over, I didn't know about https://github.com/scalacenter/scalafix/blob/d31b5dca1fdfd302256238af658f4016006f40c6/build.sbt#L209). WDYT?

bjaglin:
https://github.com/scalacenter/scalafix/pull/1275

mlachkar:
Yes definitely we can skip this release and I can do a fix that reverts this breaking change.
 
The old way of enabling semanticdb, write semanticdb files in `/classes`, and then we only needed to give the classpath to testkit, which automatically include semanticdb files.
When we enable semanticdb with the new way, files go to `/meta`.  So I had to change the classpath given to testkit to include `/meta`.
The other solution is just to add them as before, in `/classes` by doing `semantidbInJar := true`

What do you think about that? 
Another question: What will happen with the 0.9.22? Is scala steward going to upgrade the project to this version?

mlachkar:
I just saw the PR. I like this solution.

bjaglin:
>  Another question: What will happen with the 0.9.22? Is scala steward going to upgrade the project to this version?

[It has already](https://github.com/search?q=author%3Ascala-steward+is%3Apr+scalafix), but I think it's OK if we tag a 0.9.23 right away.

bjaglin:
Failures of the last run look transient. But I guess we'll close this and reopen a clean bump against  0.9.23 to confirm there is no longer a breaking change?

mlachkar:
yes. I agree

mlachkar:
We will directly move to 0.9.23

PR #181, opened 2020-12-01
support running scalafix on projects with sources at the root

bjaglin:
When defining local rules using the Scalafix configuration, libraryDependencies must be defined, as documented in: https://scalacenter.github.io/scalafix/docs/developers/local-rules.html. We could add it by default, but it would unnecessarily clutter the update task in 99.9% of the plugin use cases.

When projects have a custom directory layout (via scalaSource), users could not know that Scalafix settings would need to be updated to exclude sources potentially at the root (sourcesInBase), in order to prevent compilation errors as Scalafix does not have the Scala libs, causing:
> scala.reflect.internal.MissingRequirementError: object scala in compiler mirror not found.

This commit disables sourcesInBase by default to prevent this, as the local rules have no reason to share code with other default configurations anyway.


bjaglin:
Originally reported in https://gitter.im/scalacenter/scalafix?at=5fb8d18bba69631c748f917b

olafurpg (approved):
I'm not sure I fully understand the details of this issue but the change looks good to me 👍 I only use the `src/{main,test}/scala` convention in my projects

PR #189, opened 2021-02-23
Deterministic `args` file stamp

gontard:
Since https://github.com/scalacenter/sbt-scalafix/pull/188 the scalafix is _almost_ machine-independent but we discovered that the `args` was not determinist across machines. 
This PR fixes this issue by sorting the classpath and the dependencies before hashing them.


bjaglin (approved):

bjaglin:
I can't find an easy way to guard this against regression via a scripted. The best I could think of would be to compare the `args` stamp generated by a scripted across JVMs/OS on GitHub Actions using a shared volume somehow, but that's far from trivial... I think this is something we'll have to consider if/when we leverage/advertise support for https://github.com/sbt/sbt/issues/6298 in sbt-scalafix.

PR #195, opened 2021-03-03
scalafixEnable: relax -Xfatal-warnings for scalafix invocations

bjaglin:
Closes https://github.com/scalacenter/scalafix/issues/1346

`scalafixEnable` should maximize the chances of getting `semanticdb-scalac` to run so that semantic rules can run.

When enabled in the build, `-Xfatal-warnings` can be a major obstacle for example when:
1. `scalafixEnable` forced a newer (patch) Scala version that the one in the build, with improved warning reporting;
2. the user already has a build with fatal deprecation warnings and wants to apply a rewrite rule to fix them.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixEnable.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixEnable.scala
@@ -3,6 +3,7 @@ package scalafix.sbt
 import sbt._
 import sbt.Keys._
 import sbt.internal.sbtscalafix.Compat
+import ScalafixPlugin.autoImport._
 
 /** Command to automatically enable semanticdb-scalac for shell session */
 object ScalafixEnable {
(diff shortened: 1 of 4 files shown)


olafurpg (approved):
This is a great idea!

src/main/scala/scalafix/sbt/ScalafixEnable.scala
```diff
@@ -57,10 +69,16 @@ object ScalafixEnable {
...
 
-    val semanticdbInstalled = Compat.append(extracted, settings, s)
+    val scalafixReady = Compat.append(extracted, settings, s)
 
-    semanticdbInstalled
+    scalafixReady
   }
+
+  private def scalafixInvoked: Def.Initialize[Task[Boolean]] =
+    Def.task(executionRoots.value.exists(_.key == scalafix.key))
```
fthomas, line 83: Should this also check for `scalafixAll`?

bjaglin: good catch! just added 9c91d70

PR #196, opened 2021-03-05
Relax fatal warnings for all scalafix invocations

bjaglin:
This follows https://github.com/scalacenter/sbt-scalafix/pull/195, bringing the feature to all users of `ScalafixPlugin` (not just the ones using the `scalafixEnable` command).

The rationale is the same as the original PR: in a build with fatal warnings, when bumping a lib deprecating APIs (which provides a Scalafix migration to assist), the user shouldn't have to manually, temporarily relax `scalacOptions` to run that migration as it's cumbersome and will cause zinc invalidation locally twice (once to apply to the rule, and once to go back to fatal warnings).

This is a bit similar to (2) in https://github.com/scalacenter/bloop/pull/1001 & https://github.com/scalameta/metals/commit/ddc62ff232084a9c12fb0c4b4ad438efecd4dd3d.

This is also a step towards https://github.com/scalacenter/scalafix/issues/1199, as we might eventually add `scalacOptions` when rules requires them. However, it does not seem possible to customize a task based on the arguments of the input task invoked, so this would only work for rules declared statically in the configuration file I guess, which does not make a lot of sense as the goal was to provide one-shot users a better UX, and these users probably invoke the rule(s) as input task argument(s).

Implementation notes
- The first 2 commits make the feature less intrusive, the third one generalizes it, and the others tweak the behavior for less common scenarios. See commit messages.
- [sbt 0.13](https://github.com/scalacenter/scalafix/issues/1333) does not have the option for the incremental compiler to use the cache despite a change in `scalacOptions`, so scalafix will now trigger a recompile (without saving the analysis file though, so further explicit compiles will remain incremental).
- The [introduction of `ignoredScalacOptions` in sbt 1.2.0-M1](https://github.com/sbt/zinc/commit/8681829) means that users with 1.0.x or 1.1.x will get linking errors (sbt-scalafix is currently built [against sbt 1.2.1](https://github.com/scalacenter/sbt-scalafix/blob/1eeafe2df0c86a3cf4d90137cc3d2c00adb8b407/build.sbt#L53-L59)). As mentioned in the last commit, I don't think it's worth supporting them as the upgrade path between 1.x release is very easy anyway, and we don't test for these versions anyway so they might already be broken.


.github/workflows/ci.yml
```diff
@@ -17,6 +17,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: olafurpg/setup-scala@v10
+      - if: ${{ matrix.command == '++2.10.7 test ''scripted sbt-scalafix/* skip-windows/*''' }}
+        run: find . -regex .*relax-scalacOptions/test -exec sed -i '/> checkLastCompilationCached/d' {} \;
```
bjaglin, line 21: not particularly proud of this, but it won't stay long https://github.com/scalacenter/scalafix/issues/1333

bjaglin:
@fthomas this should be a no-op for https://github.com/scalacenter/scalafix/issues/1346, but I am not sure our scripted suite is exhaustive, so you might want to give this PR a spin.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -97,6 +97,25 @@ object ScalafixPlugin extends AutoPlugin {
     def scalafixConfigSettings(config: Configuration): Seq[Def.Setting[_]] =
       inConfig(config)(
         Seq(
+          scalacOptions := {
+            val options = scalacOptions.value
+            if (!scalafixInvokedAlone.value) options
+            else
+              options.filterNot { option =>
+                scalacOptionsToRelax.exists(_.matcher(option).matches)
+              }
```
adpi2, line 106: This `scalafixInvokedAlone` can be hard to debug when you don't know it exists. I think it is very surprising that the `scalafix` task does not behave the same when it is invoked alone or not.

Why not scoping the filtered `scalacOptions` to the `scalafix` task?
```
scalafix / scalacOptions := scalacOptions.value.filterNot(...)
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -97,6 +97,25 @@ object ScalafixPlugin extends AutoPlugin {
...
+          },
+          incOptions := {
+            val options = incOptions.value
+            if (!scalafixInvokedAlone.value) options
+            else
+              // maximize chance to get a zinc cache hit when running scalafix, even though we have
+              // potentially added/removed scalacOptions for that specific invocation
+              Compat.addIgnoredScalacOptions(
+                options,
+                scalacOptionsToRelax.map(_.pattern())
```
adpi2, line 116: same comment here

fthomas:
Just checked ths PR and it works for me too.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: Thanks for the review @adpi2 !

> I think it is very surprising that the scalafix task does not behave the same when it is invoked alone or not.

Indeed. I guess we could just fail fast if we detect other tasks? As described in the notes of https://github.com/scalacenter/sbt-scalafix/pull/196/commits/abb7e1bb3504f3ff81e1bb71c0331293c2540812, `scalafix[All]` being an input task, it is a very unlikely scenario anyway. I just added cdffae54a5e4585c5a5770b5402e38fc4fc85f62.

> Why not scoping the filtered scalacOptions to the scalafix task?

The problem is that we want to affect the regular Zinc compilation which does not look up `scalafix / scalacOptions`.

1. When semantic rules are requested, `scalafix` needs classes and semanticdb information to run.
1. It uses the [regular compilation output directory](https://github.com/scalacenter/sbt-scalafix/blob/1eeafe2df0c86a3cf4d90137cc3d2c00adb8b407/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L443) (as opposed to a separate compilation) to leverage the incremental compilations requested by the user in their normal workflow.
1. [An explicit call to `compileIncremental` (via `manipulateBytecode`)](https://github.com/scalacenter/sbt-scalafix/blob/1eeafe2df0c86a3cf4d90137cc3d2c00adb8b407/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L453-L454) is done, to make sure that the compiler output is up to date, failing before calling Scalafix if the compilation is not successful.
1. To avoid getting warning as errors that would prevent generation of class and semanticdb files and defeat the purpose of Scalafix migration rules, this PR alters the `compileIncremental`, transitively invoked by `scalafix`.

To follow you suggestion while still honoring the goals of (2) & (4), I tried to define a separate, more-relaxed `compileIncremental`:
```scala
// pseudo code showing that relaxed compilation would only be triggered in case of failure of the regular one
scalafix := scalafixInputTaskImpl
  .dependsOn(compileIncremental.recoverWith(scalafix / compileIncremental)).value

inTask(scalafix)(
  Seq(
    scalacOptions -= "-Xfatal-warnings",
    compileIncremental := compileIncrementalTask,
  ) ++ compileInputsSettings // compileInputs looks up compileOptions which looks up scalacOptions
)
```
But the problem is that [`compileInputs` is explicitly looked up within `compile`](https://github.com/sbt/sbt/blob/f155bfd1f3bdcea4b66c0480c1cef79b74410559/main/src/main/scala/sbt/Defaults.scala#L2262), so the enclosing `inTask(scalafix)` has no effect where we would like it to: `scalafix / compileInputs` and thus `scalafix / scalacOptions` are not looked up.

Any idea?

adpi2: Thanks for the explanation, it was not clear to me that the `scalafix` task could trigger a compilation through the uncommon dependency on `manipulateBytecode`.

So you are right, the `scalafix / compileInputs` are not looked up by the regular `compileIncremental` task and I don't see any solution.

The fail-fast strategy makes things a bit better, but I am still worried about the current implementation which is quite convoluted. Also I have been able to reproduce some bug while testing:

1. In a project that contains at least 2 sources, the top one depending on the bottom one.
2. In the top source, add a piece of code that produces a warning
```
def foo {}
```
3. Run compile, it fails
```
sbt:root> compile
[info] compiling 1 Scala source to /home/piquerez/scalacenter/sbt-scalafix-196/target/scala-2.13/classes ...
[error] /home/piquerez/scalacenter/sbt-scalafix-196/src/main/scala/foo/Main.scala:5:11: procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `foo`'s return type
[error]   def foo {}
[error]           ^
[error] No warnings can be incurred under -Werror.
[error] two errors found
[error] (Compile / compileIncremental) Compilation failed
[error] Total time: 3 s, completed Mar 8, 2021, 11:04:47 AM
```
4. Run scalafixAll, it should compile one source successfully
```
sbt:root> scalafixAll
[info] compiling 1 Scala source to /home/piquerez/scalacenter/sbt-scalafix-196/target/scala-2.13/classes ...
[warn] /home/piquerez/scalacenter/sbt-scalafix-196/src/main/scala/foo/Main.scala:5:11: procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `foo`'s return type
[warn]   def foo {}
[warn]           ^
[warn] one warning found
[info] Running scalafix on 2 Scala sources
[success] Total time: 1 s, completed Mar 8, 2021, 11:05:41 AM
```
5. Run compile again
```
sbt:root> compile
[info] compiling 2 Scala sources to /home/piquerez/scalacenter/sbt-scalafix-196/target/scala-2.13/classes ...
[error] /home/piquerez/scalacenter/sbt-scalafix-196/src/main/scala/foo/Main.scala:5:11: procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `foo`'s return type
[error]   def foo {}
[error]           ^
[error] No warnings can be incurred under -Werror.
[error] two errors found
[error] (Compile / compileIncremental) Compilation failed
[error] Total time: 0 s, completed Mar 8, 2021, 11:06:35 AM
```
The 2 sources are being recompiled which means that the Zinc cache has been completely invalidated, for no reason. In a bigger project all files are being recompiled.

adpi2: An alternative solution could be to introduce a `scalafixCompile` task that duplicate the `compile` logic using the same Zinc cache and the same semanticdb target to leverage the incremental compilation.

The benefits would be:
1. Break the cyclic dependency when `scalaOnCompile := true`
2. Make the relax scalacOptions implementation more simple: `compile` depends on `compile / scalacOptions` while `scalafixCompile` depends on `scalafixCompile / scalacOptions`

But also it has some drawbacks:
1. Can be hard to implement: in particular reusing the Zinc cache without altering it
2. Duplication of code: hard to maintain if the sbt implementation evolve

@bjaglin I let you decide if this alternative is worth trying or not. Otherwise I would be okay to validate the current solution.

bjaglin: > Also I have been able to reproduce some bug while testing

Good catch. I just added 8f407e1aeb7bf994f290138de5d2e7c19355da79 which should cover this. I couldn't find a simple way to assert this behavior in the scripted, so I made sure the zinc analysis file was not saved for scalafix invocations, which should prevent other related bugs.

bjaglin: > An alternative solution could be to introduce a scalafixCompile task that duplicate the compile logic using the same Zinc cache and the same semanticdb target to leverage the incremental compilation.

I initially ruled out that option as I thought it was would be too complicated, and unfortunately, after digging a bit more into it, I am not very hopeful.
1. As you mention in drawback (2), it would be very hard to track sbt changes and support a decent range of sbt versions, just looking at how `compileIncrementalTaskImpl` has evolved in the last months with BSP & pipeling. This is a case for [derived settings](https://github.com/sbt/sbt/discussions/6323) I believe.
1. Even though the current solution can be confusing as it changes the behavior of the Zinc compilation based on the invoked task, checking the logs of `compileIncremental` after a scalafix run does show a bit what's going on. The output of a potential `scalafixCompileIncremental` (classes & semantidcb files) would be shared with the regular `compileIncremental`, making it maybe a bit more obscure and subject to races.
1. Apart from relaxing the scalac invocation in the scope(s) where scalafix was specifically invoked, parent configs/projects (via `dependencyClasspath`) should probably also be relaxed like the current implementation of this PR does. Running a migration script fixing warnings would require many successive runs if we fail to do so, defeating the UX goal of this PR.

bjaglin: Yet another approach to this PR would be to be resilient to a failing `compileIncremental` and start scalafix anyway against stale/missing class files, relying on a [on-demand, relaxed compilation in scalafix for semanticdb info](https://github.com/scalacenter/scalafix/blob/fd1ceef8f2a20592e79eae743bef3c02eadccad6/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala#L338). At the moment the `SymbolTable` used by scalafix does not make use of that on-demand compilation though (and I don't know if we even can?), so this would be very hacky.

bjaglin: @adpi2 happy to hear your thoughts in case I missed something

bjaglin:
@olafurpg any take on the feature and/or the implementation? Thanks!

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -97,6 +97,25 @@ object ScalafixPlugin extends AutoPlugin {
     def scalafixConfigSettings(config: Configuration): Seq[Def.Setting[_]] =
       inConfig(config)(
         Seq(
+          scalacOptions := {
```
olafurpg: What happens in builds that do `scalacOptions :=` overriding the settings from plugins? My experience is that quite many builds do this with `scalacOptions`.

olafurpg (commented):
I think this is a really worthwhile problem to solve because many users want to enable fatal warnings together with Scalafix. I am a bit on the fence about this particular implementation because users may end invalidating the compile cache for every compile/scalafix cycle.

My default workflow with Scalafix is to iterate on `~test` and then run `scalafix` before opening a PR. It would be a regression if the `scalafix` invocation required a clean re-compile of the build.

(just thinking out loud) One possible alternative approach would be to change the sbt reporter to re-implement fatal warnings. This might requires changes to scalac itself so that it continues compilation even when warnings are promoted to errors. This is probably more ambitious than what you had in mind, but I suspect you could support from upstream sbt/scalac maintainers to implement this. I think it's also worthwhile to consider whether Scala 3 should support something like this. cc/ @eed3si9n

olafurpg:
Regardless of my comment, feel free to merge this feature! We should not let perfect become the enemy of good :) The test suite in this PR looks very well done and the current solution does solve a big problem that many Scalafix users face today.

bjaglin:
> (just thinking out loud) One possible alternative approach would be to change the sbt reporter to re-implement fatal warnings. This might requires changes to scalac itself so that it continues compilation even when warnings are promoted to errors. This is probably more ambitious than what you had in mind, but I suspect you could support from upstream sbt/scalac maintainers to implement this. I think it's also worthwhile to consider whether Scala 3 should support something like this. cc/ @eed3si9n

Bloop does something like this, [taking ownership of fatal warnings](https://github.com/scalacenter/bloop/pull/1001). I guess it would be useful to have that upstream somehow, but it does look like a significant amount of work.

olafurpg:
Yeah, I was gonna mention that Bloop does something like this. However, Bloop only doesn't do it downstream projects in a multi-module build. An ideal solution would complete the full multi-project compilation as if fatal warnings were disabled but then fail the command with a non-zero exit code.

olafurpg:
There are UX challenges even if you go that route, it might be confusing for users to see the compilation continue despite compile errors. I'm not sure what the best solution 🤷 I don't use fatal warnings personally. If I were to use them in my project then I would only enable them in CI.

bjaglin:
>  I am a bit on the fence about this particular implementation because users may end invalidating the compile cache for every compile/scalafix cycle.
> 
> My default workflow with Scalafix is to iterate on `~test` and then run `scalafix` before opening a PR. It would be a regression if the `scalafix` invocation required a clean re-compile of the build.

Agreed. https://github.com/scalacenter/sbt-scalafix/pull/196/commits/8f407e1aeb7bf994f290138de5d2e7c19355da79#diff-06169c9b17ca1913e8659adfb6e67d1360dcfadd638128457e2d74a9e4dcc72aR679 does make sure that analysis files are not stored for relaxed compilation triggered by scalafix, thanks to the use-case raised by @adpi2, which makes the current implementation much more robust. My assumption is that only sources that had fatal warnings (and thus for which a compilation is necessary anyway) would produce fresh class files, so the invalidation scope of the next regular incremental compilation would be limited to those. As far as I can tell, that means that that there should be no invalidation at all if there was no fatal warning (either because the flag is not enabled or because there was no warning). I'd be happy to add more scenarios to the scripted but I am not sure which ones.

eed3si9n:
> One possible alternative approach would be to change the sbt reporter to re-implement fatal warnings.

+1. It might be relatively simple to check for warning counts at the end of `compile` task and raise and error at the end on sbt side. If we strip out `-Xfatal-warnings`, compiler wouldn't have to know. The UX would be different since it would show `[warning]` rather than `[error]` but that might be an improvement.

olafurpg:
@eed3si9n thank you for pitching in! It may get bit a more complicated with -Werror in 2.13 because sbt won't know what warnings to promote to errors. I'm not sure if it requires changes on the compiler side to handle that case. Either way, such an approach is probably beyond the scope of this PR.

> does make sure that analysis files are not stored for relaxed compilation triggered by scalafix,

@bjaglin That is really cool. I'm excited to try this out. I didn't dig yet into the implementation details, it looks like you have already covered the several important situations.

bjaglin:
>  It may get bit a more complicated with -Werror in 2.13 because sbt won't know what warnings to promote to errors

I guess you mean `-Wconf`? Indeed, and that has been backported to 2.12 AFAIK. It's gaining so much traction that it would be a shame to "only" support the binary `-Xfatal-warnings`

> Either way, such an approach is probably beyond the scope of this PR.

I wonder if this is something the Scalacenter could drive, since I believe it would be beneficial for Metals as well?

I think I am going to go ahead and merge this start using the resulting SNAPSHOT in our internal build right away to look for corner cases. Any objection @mlachkar ?

mlachkar:
@bjaglin I agree. Let's merge it! Thanks for driving this change!

PR #200, opened 2021-03-25
fix following the rename of master branch

mlachkar:


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
bjaglin, line 1: add `.vscode` and `.metals` to `.gitignore` ?

mlachkar: yesss I should !

PR #210, opened 2021-05-10
initial support for running rules on Scala 3 sources

bjaglin:
Showcases https://github.com/scalacenter/scalafix/pull/1392

See open questions / limitations in https://github.com/scalacenter/scalafix/issues/1316#issuecomment-839074644

**Note that most of the [built-in Scalafix rules](https://scalacenter.github.io/scalafix/docs/rules/overview.html) do not support Scala 3 sources at the moment.**

PR #211, opened 2021-05-12
make it possible to and do build against a scalafix SNAPSHOT

bjaglin:
Follows https://github.com/scalacenter/sbt-scalafix/pull/210

It's useful to build and run all scripted tests against a scalafix SNAPSHOT to avoid getting (late) surprises during the 2-step (scalafix/sbt-scalafix) release phase.

Most of the users running SNAPSHOT versions of sbt-scalafix should be able to use one that depends on a SNAPSHOT scalafix since they must already have sonatype snapshots in the metabuild and the default `scalafixResolvers` will re-resolve scalafix through sonatype snapshots via the embedded coursier.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixTestkitPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixTestkitPlugin.scala
@@ -27,6 +27,11 @@ object ScalafixTestkitPlugin extends AutoPlugin {
   }
   import autoImport._
 
+  override def buildSettings: Seq[Def.Setting[_]] =
+    List(
+      resolvers += Resolver.sonatypeRepo("snapshots")
+    )
+
   override def projectSettings: Seq[Def.Setting[_]] =
     List(
       scalafixTestkitInputScalacOptions := scalacOptions.value,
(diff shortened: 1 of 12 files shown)


src/sbt-test/sbt-1.3/testkit/tests/src/test/scala/fix/InputOutputSuite.scala
```diff
@@ -2,15 +2,16 @@ package fix
...
 object IntputOutputSuite {
   def main(args: Array[String]): Unit = {
     if (Array("--save-expect").sameElements(args)) {
-      val suite = new SemanticRuleSuite(
-        TestkitProperties.loadFromResources(),
-        isSaveExpect = true
-      ) {
+      val suite = new AbstractSemanticRuleSuite with FunSuiteLike {
+        override val props = TestkitProperties.loadFromResources()
+        override val isSaveExpect = true
```
bjaglin, line 13: Non-trivial usage of `SemanticRuleSuite` that https://github.com/scalacenter/scalafix/pull/1395/files won't cover. See also https://github.com/scalacenter/scalafix/issues/1147.

src/main/scala/scalafix/sbt/ScalafixTestkitPlugin.scala
```diff
@@ -27,6 +27,11 @@ object ScalafixTestkitPlugin extends AutoPlugin {
   }
   import autoImport._
 
+  override def buildSettings: Seq[Def.Setting[_]] =
+    List(
+      resolvers += Resolver.sonatypeRepo("snapshots")
```
bjaglin, line 32: users of a SNAPSHOT testkit plugin must resolve from sonatype snapshots both scalafix-testkit for the test project (where the plugin is enabled) AND scalafix-core for the actual rule project, thus the need for impacting the entire build and not just the project settings.

PR #219, opened 2021-05-16
Support Scala 3 in ScalafixTestkitPlugin

bjaglin:
Integrating iterations of https://github.com/scalacenter/scalafix.g8/pull/19 for a proper way to test against different scala versions.

This effectively drops support of sbt 0.13 in `ScalafixTestkitPlugin`, since it now depends on sbt-projectmatrix which is not available.

PR #220, opened 2021-05-17
ScalafixTestkitPlugin improvements

bjaglin:
Follows https://github.com/scalacenter/sbt-scalafix/pull/219

### DIFF
--- a/src/main/scala-sbt-1.0/scalafix/sbt/ScalafixTestkitPlugin.scala
+++ b/src/main/scala-sbt-1.0/scalafix/sbt/ScalafixTestkitPlugin.scala
@@ -13,38 +13,42 @@ object ScalafixTestkitPlugin extends AutoPlugin {
   override def requires: Plugins = JvmPlugin
 
   object autoImport {
-    case class InputAxis(scalaVersion: String) extends VirtualAxis.WeakAxis {
+    case class TestkitTargetAxis(scalaVersion: String)
+        extends VirtualAxis.WeakAxis {
       private val scalaBinaryVersion =
         CrossVersion.binaryScalaVersion(scalaVersion)
 
-      override val idSuffix = s"Input${scalaBinaryVersion.replace('.', '_')}"
-      override val directorySuffix = s"input$scalaBinaryVersion"
+      override val idSuffix = s"Target${scalaBinaryVersion.replace('.', '_')}"
+      override val directorySuffix = s"target$scalaBinaryVersion"
     }
 
-    object InputAxis {
-      def inputScalaVersion(virtualAxes: Seq[VirtualAxis]): String =
-        virtualAxes.collectFirst { case a: InputAxis => a.scalaVersion }.get
+    object TestkitTargetAxis {
+
+      private def targetScalaVersion(virtualAxes: Seq[VirtualAxis]): String =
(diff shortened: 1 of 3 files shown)


src/sbt-test/sbt-1.5/testkit/build.sbt
```diff
@@ -38,40 +38,41 @@ lazy val tests = projectMatrix
   .settings(
     libraryDependencies += "ch.epfl.scala" % "scalafix-testkit" % V.scalafixVersion % Test cross CrossVersion.full,
     scalafixTestkitOutputSourceDirectories :=
-      resolveByInputAxis(output, Compile / unmanagedSourceDirectories).value,
```
bjaglin, line 41: https://github.com/scalacenter/scalafix.g8/pull/19#discussion_r633569300

src/main/scala-sbt-1.0/scalafix/sbt/ScalafixTestkitPlugin.scala
```diff
@@ -13,38 +13,42 @@ object ScalafixTestkitPlugin extends AutoPlugin {
   override def requires: Plugins = JvmPlugin
 
   object autoImport {
-    case class InputAxis(scalaVersion: String) extends VirtualAxis.WeakAxis {
+    case class TestkitTargetAxis(scalaVersion: String)
```
bjaglin, line 16: not very happy about that name either... suggestions welcome!

src/main/scala-sbt-1.0/scalafix/sbt/ScalafixTestkitPlugin.scala
```diff
@@ -13,38 +13,42 @@ object ScalafixTestkitPlugin extends AutoPlugin {
...
-    def resolveByInputAxis[T](
-        matrix: ProjectMatrix,
-        key: TaskKey[T]
-    ): Def.Initialize[Task[T]] =
-      Def.taskDyn {
-        val sv = InputAxis.inputScalaVersion(virtualAxes.value)
-        val project = matrix.finder().apply(sv)
-        Def.task((project / key).value)
-      }
+      private def targetScalaVersion(virtualAxes: Seq[VirtualAxis]): String =
```
mlachkar, line 27: I don't completely get why we take the first scalaVersion that we find, and use it the resolve task

src/main/scala-sbt-1.0/scalafix/sbt/ScalafixTestkitPlugin.scala
mlachkar: It's already better! Don't have better propositions

mlachkar (approved):
LGTM

src/main/scala-sbt-1.0/scalafix/sbt/ScalafixTestkitPlugin.scala
bjaglin: > why we take the first scalaVersion that we find

because there can be only one value per axis per project

> use it the resolve task

we lookup the corresponding input/output project based on this value, that's where I am not too satisfied with the `Target` naming: it's about tagging in `tests*` which `input*` or `output*` it should run against.

PR #225, opened 2021-05-28
drop support for sbt 0.13

bjaglin:
Closes https://github.com/scalacenter/scalafix/issues/1333

sbt 0.13 users will still be able to get the latest version of Scalafix by explicitly including it:
```diff
 // project/plugins.sbt
 addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.29") // latest version published for sbt 0.13
+dependencyOverrides += "ch.epfl.scala" % "scalafix-interfaces" % "0.9.x"
```


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -477,8 +473,19 @@ object ScalafixPlugin extends AutoPlugin {
...
+      val argWriter = new JsonWriter[Arg.CacheKey] {
+
+        override final def write[J](
+            obj: Arg.CacheKey,
+            builder: Builder[J]
+        ): Unit = stamp(obj)(builder)
+
+        private def stamp[J](
+            obj: Arg.CacheKey
+        )(implicit builder: Builder[J]): Unit = obj match {
```
bjaglin, line 488: There is way too much logic in `ScalafixPlugin.scala`. I'll try to extract that in a separate PR.

bjaglin:
any objection @olafurpg ?

olafurpg:
No objections from me. If it’s easy to support, it might be worth releasing a 0.13 version that support loading a configurable version of Scalafix. That would allow the latest 0.13 version to call future versions of Scalafix. But I wouldn’t worry about it if it’s too much effort.

bjaglin:
> If it’s easy to support, it might be worth releasing a 0.13 version that support loading a configurable version of Scalafix.

That was my initial though too, and I have been postponing this until I get free time to tackle https://github.com/scalacenter/scalafix/issues/1146, but the last  PRs showed that it was getting increasingly annoying, so I got tired of it. Since getting a scalafix version is as simple as https://scalacenter.github.io/scalafix/docs/users/installation.html#snapshot, I assume we can write up something if there is interest.

I remember you posting stats about Sonatype downloads at some point, is that something you still have access to? I guess it's not very accurate but it could give an idea of the current sbt 0.13 user base.

olafurpg:
That workaround is good enough (exactly what I was looking for). 

Sonatype doesn’t provide download stats on sbt plugins unfortunately. I think it has something to do with how sbt plugins artifacts use Ivy patterns.

bjaglin:
Not sure why this got closed after my rebase against https://github.com/scalacenter/sbt-scalafix/pull/228 :thinking:

Following up in https://github.com/scalacenter/sbt-scalafix/pull/230.

I have added a note in https://github.com/scalacenter/scalafix/pull/1417/files#diff-2e28f911279567cbb268785581f61db7f687f49d4adedbd85a06b0943fbcd320 for the best-effort sbt 0.13.x support.

PR #241, opened 2021-06-30
Disable BSP in Scalafix config

adpi2:
The Scalafix config adds a BSP build target that is useless because it has no sources.

This PR disables BSP in the Scalafix config so that it is never exported to the IDE. The implementation involves a bit of reflection to deal with the compatibility issue on old versions of sbt.

See also:
- https://github.com/sbt/sbt/issues/6546
- https://github.com/scalameta/metals/pull/2895

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -183,15 +183,20 @@ object ScalafixPlugin extends AutoPlugin {
   override lazy val projectConfigurations: Seq[Configuration] =
     Seq(ScalafixConfig)
 
-  override lazy val projectSettings: Seq[Def.Setting[_]] =
-    Seq(Compile, Test).flatMap(c => inConfig(c)(scalafixConfigSettings(c))) ++
-      inConfig(ScalafixConfig)(
-        Defaults.configSettings :+ (sourcesInBase := false)
-      ) ++
-      Seq(
-        ivyConfigurations += ScalafixConfig,
-        scalafixAll := scalafixAllInputTask.evaluated
+  override lazy val projectSettings: Seq[Def.Setting[_]] = Def.settings(
+    Seq(Compile, Test).flatMap(c => inConfig(c)(scalafixConfigSettings(c))),
+    inConfig(ScalafixConfig)(
+      Def.settings(
+        Defaults.configSettings,
+        sourcesInBase := false,
+        // local copy of https://github.com/sbt/sbt/blob/e4231ac03903e174bc9975ee00d34064a1d1f373/main/src/main/scala/sbt/Keys.scala#L400
+        // so that it does not break on sbt version below 1.4.0
+        SettingKey[Boolean]("bspEnabled") := false
       )
+    ),
+    ivyConfigurations += ScalafixConfig,
+    scalafixAll := scalafixAllInputTask.evaluated
+  )
 
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
     scalafixConfig := None, // let scalafix-cli try to infer $CWD/.scalafix.conf


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -183,15 +183,20 @@ object ScalafixPlugin extends AutoPlugin {
...
 
-  override lazy val projectSettings: Seq[Def.Setting[_]] =
-    Seq(Compile, Test).flatMap(c => inConfig(c)(scalafixConfigSettings(c))) ++
-      inConfig(ScalafixConfig)(
-        Defaults.configSettings :+ (sourcesInBase := false)
-      ) ++
-      Seq(
-        ivyConfigurations += ScalafixConfig,
-        scalafixAll := scalafixAllInputTask.evaluated
+  override lazy val projectSettings: Seq[Def.Setting[_]] = Def.settings(
```
bjaglin, line 186: nice, I didn't know that trick

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -728,4 +733,21 @@ object ScalafixPlugin extends AutoPlugin {
...
+    * deal with the binary compatibility of different versions of sbt.
+    *
+    * For instance the bspEnabled key has been introduced in sbt 1.4. Calling
+    * tryLoadSbtSetting("bspEnabled") return None on sbt 1.3 and
+    * Some(Keys.bspEnabled) on sbt 1.4 and greater.
+    */
+  private def tryLoadSbtSetting[T](key: String): Option[SettingKey[T]] = {
+    try {
+      val getter = Keys.getClass().getMethod(key)
+      if (getter == null) println("key is null")
```
bjaglin, line 747:
```suggestion
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -728,4 +733,21 @@ object ScalafixPlugin extends AutoPlugin {
...
     List("-Xfatal-warnings", "-Werror", "-Wconf.*").map(_.r.pattern)
+
+  /** Try load a setting key from sbt.Keys using reflection. This is usefull to
+    * deal with the binary compatibility of different versions of sbt.
+    *
+    * For instance the bspEnabled key has been introduced in sbt 1.4. Calling
+    * tryLoadSbtSetting("bspEnabled") return None on sbt 1.3 and
+    * Some(Keys.bspEnabled) on sbt 1.4 and greater.
+    */
+  private def tryLoadSbtSetting[T](key: String): Option[SettingKey[T]] = {
```
bjaglin, line 744: this will be handy! I wonder though if wouldn't it be simpler to declare a local `bspEnabled` key with the same type as the upstream one?

bjaglin (commented):
Thanks for the PR! Looking back, introducing `ScalafixConfig` to allow definition of local rules was a bad decision, so I am hoping we can deprecate it soon and replace it with a more adhoc mechanism, so this kind of problems wouldn't show up.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
adpi2: I have tried that:
```scala
        settingKey[Boolean]("bspEnabled") := false,
```
but it did not work.

Do you know if there is a working example somewhere?

adpi2: I mean the original `bspEnabled` setting is still true after that.

bjaglin: > Do you know if there is a working example somewhere?

I had good results copying https://github.com/scalacenter/bloop/blob/v1.4.0/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala#L76-L77 in a private project for a somewhat similar use-case. It just worked in that case, but `bspEnabled` seems to be a non-regular key in the way it's consumed?

adpi2: Indeed your solution does work.

> bspEnabled seems to be a non-regular key in the way it's consumed?

I would say it is a regular key, but `Global / bspEnabled` is consumed very early by sbt server itself.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -53,6 +53,10 @@ object ScalafixPlugin extends AutoPlugin {
         "Make scalafix invocations incremental. On by default."
       )
 
+    // copy of https://github.com/sbt/sbt/blob/e4231ac03903e174bc9975ee00d34064a1d1f373/main/src/main/scala/sbt/Keys.scala#L400
+    // so that it is available on sbt versions older than 1.4.0
+    val bspEnabled = settingKey[Boolean]("")
```
bjaglin, line 58: Could you move it out of`autoImport` so it does not shadow the real one in `.sbt` files? (maybe the order of the imports would handle that, but just in case, for IDEs "go to definition" for example)

bjaglin:
> The Scalafix config adds a BSP build target that is useless because it has no sources.

For the record, this is not true for all users, see https://scalacenter.github.io/scalafix/docs/developers/local-rules.html#single-project-sbt-build. However, it's a very niche feature that adds a lot much overhead (a custom config) for 0,01% of the sbt-scalafix users, so to elaborate what I wrote earlier, I'd like to remove this feature and only support https://scalacenter.github.io/scalafix/docs/developers/local-rules.html#as-a-separate-sub-project, which would effectively remove the need for that ivy config.

adpi2:
> For the record, this is not true for all users, see https://scalacenter.github.io/scalafix/docs/developers/local-rules.html#single-project-sbt-build.

In the case described in the documentation, having BSP enabled would be useful. But as you said it is kind of a niche feature and I agree it would better to use a new project instead.

bjaglin (approved):

PR #261, opened 2021-11-26
Release scalafix 0.9.33

mlachkar:


mlachkar:
Java 17 tests are failing with this error. I have tried to update the tests to make sure to use only the latest scalaVersions taht support java 17, but it doesn't fix the tests.
I have test manually, everything seems to work, so I m merging and trying to find how to fix those tests later.

```
[error] error: error while loading String, class file '/modules/java.base/java/lang/String.class' is broken
[error] (class java.lang.NullPointerException/Cannot invoke "scala.tools.nsc.Global$Run.typerPhase()" because the return value of "scala.tools.nsc.Global.currentRun()" is null)
```

PR #262, opened 2021-11-30
Add more fine grained check for `scalacOptionsToRelax`

ckipp01:
Creating this to start a conversation about it and to see if this would be alright as a follow-up to some of the work done in https://github.com/scalacenter/sbt-scalafix/pull/196. If I understood the conversation right you want to avoid situations where this is set to error out causing issues. However, currently it strips out any usage of `-Wconf`. The situation that I'm in is that we want to avoid warnings from generated sources since we generate a _ton_ of files with scalaxb that contains a whole bunch of compiler warnings.  To silence these we use `-Wconf:src=src_managed/.*:silent`. This works great during normal compilation, but then when we run scalafix check in our CI our logs end up being full or warnings causing a scare. Would it be possible to make this more fine-grained like this to specifically target `:e` or `:error`?

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -732,5 +732,5 @@ object ScalafixPlugin extends AutoPlugin {
     }
 
   private val scalacOptionsToRelax =
-    List("-Xfatal-warnings", "-Werror", "-Wconf.*").map(_.r.pattern)
+    List("-Xfatal-warnings", "-Werror", "-Wconf.*:(e|error)").map(_.r.pattern)
 }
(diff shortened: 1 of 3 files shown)


bjaglin:
> Would it be possible to make this more fine-grained like this to specifically target :e or :error?

Makes sense :+1:

Could you reproduce your scenario by adding a (mandatory) silencing wconf in https://github.com/scalacenter/sbt-scalafix/blob/f6fffde77e5ba520db81f75984a5510444391931/src/sbt-test/sbt-scalafix/relax-scalacOptions/build.sbt#L14? Thanks!

bjaglin:
It looks like the CI failure is unrelated by the way, see https://github.com/scalacenter/sbt-scalafix/pull/261#issuecomment-979977277.

ckipp01:
> Could you reproduce your scenario by adding a (mandatory) silencing wconf in

I _think_ what I have should work now.

bjaglin:
Thanks! You can use a snapshot once master is built: https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/sbt-scalafix_2.12_1.0/

PR #292, opened 2022-03-24
make scalafixEnable less aggressive & more future-proof

bjaglin:
1. Keep the build scala version when semanticdb-scalac is available for that scala version for the scalameta version that scalafix tracks.
1. If not, try to find the earliest binany-compatible semanticdb-scalac (preferred over the latest last to be more repeatable) and use it instead.
1. Only when that fails (or resolution fails), force the scala version as previously.

(1) avoids unnecessary scala bumps which can result in
- zinc cache invalidation
- compilation failures because of stricter compiler checks
- resolution failures because of yet-unpublished compiler plugins (fixes https://github.com/scalacenter/scalafix/issues/1573)

(2) opens the possibility for using sbt-scalafix on scala versions released "in the future" without any scalameta/scalafix bump (as long as scalameta is published obviously).

---

```
sbt:sbt-scalafix> scripted sbt-scalafix/scalafixEnable
...
[warn] Forcing scalaVersion 2.11.12 in project scala211_old since no semanticdb-scalac version binary-compatible with 4.4.10 and cross-published for scala 2.11.0 was found - consider bumping scala
[info] Using semanticdb-scalac 4.1.0 in project scala211 since the version 4.4.10 tracked by scalafix 0.9.27 is not available for scala 2.11.11 - consider upgrading sbt-scalafix or bumping scala
[info] Using semanticdb-scalac 4.4.28 in project scala212 since the version 4.4.10 tracked by scalafix 0.9.27 is not available for scala 2.12.15 - consider upgrading sbt-scalafix or bumping scala
```

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixEnable.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixEnable.scala
@@ -116,7 +116,7 @@ object ScalafixEnable {
                   scalaVersion := {
                     val v = recommendedSemanticdbScalacScalaV.toString
                     sLog.value.warn(
-                      s"Forcing scalaVersion $v in project " +
+                      s"Forcing scalaVersion to $v in project " +
                         s"${project.ref.project} since no semanticdb-scalac " +
                         s"version binary-compatible with $recommendedSemanticdbV " +
                         s"and cross-published for scala " +
@@ -125,26 +125,23 @@ object ScalafixEnable {
                     )
                     v
                   },
-                  semanticdbCompilerPlugin :=
-                    semanticdbCompilerPlugin.value
-                      .withRevision(recommendedSemanticdbV.toString)
+                  semanticdbVersion := recommendedSemanticdbV.toString
                 )
               case Success(available)
                   if available.contains(recommendedSemanticdbV) =>
                 Seq(
-                  semanticdbCompilerPlugin :=
-                    semanticdbCompilerPlugin.value
-                      .withRevision(recommendedSemanticdbV.toString)
+                  semanticdbVersion := recommendedSemanticdbV.toString
                 )
               case Success(earliestAvailable :: tail) =>
                 val futureVersion =
                   SemanticSelector.apply(s">${recommendedSemanticdbV}")
+
                 if (earliestAvailable.matchesSemVer(futureVersion)) {
                   Seq(
-                    semanticdbCompilerPlugin := {
+                    semanticdbVersion := {
                       val v = earliestAvailable.toString
                       sLog.value.info(
-                        s"Using semanticdb-scalac $v in project " +
+                        s"Setting semanticdbVersion to $v in project " +
                           s"${project.ref.project} since the version " +
                           s"${recommendedSemanticdbV} tracked by scalafix " +
                           s"${BuildInfo.scalafixVersion} will not be " +
@@ -152,30 +149,36 @@ object ScalafixEnable {
                           s"${project.scalaVersion0.toString} - " +
                           s"consider upgrading sbt-scalafix"
                       )
-                      semanticdbCompilerPlugin.value.withRevision(v)
+                      v
                     }
                   )
                 } else {
                   val latestAvailable =
                     tail.lastOption.getOrElse(earliestAvailable)
-                  Seq(
-                    semanticdbCompilerPlugin := {
-                      val v = latestAvailable.toString
-                      sLog.value.info(
-                        s"Using semanticdb-scalac $v in project " +
-                          s"${project.ref.project} since the version " +
-                          s"${recommendedSemanticdbV} tracked by scalafix " +
-                          s"${BuildInfo.scalafixVersion} is no longer " +
-                          s"published for scala " +
-                          s"${project.scalaVersion0.toString} - " +
-                          s"consider bumping scala"
-                      )
-                      semanticdbCompilerPlugin.value.withRevision(v)
-                    }
-                  )
+                    Seq(
+                      semanticdbVersion := {
+                        val v = latestAvailable.toString
+                        sLog.value.info(
+                          s"Setting semanticdbVersion to $v in project " +
+                            s"${project.ref.project} since the version " +
+                            s"${recommendedSemanticdbV} tracked by scalafix " +
+                            s"${BuildInfo.scalafixVersion} is no longer " +
+                            s"published for scala " +
+                            s"${project.scalaVersion0.toString} - " +
+                            s"consider bumping scala"
+                        )
+                        v
+                      }
+                    )
                 }
             }
-          } :+ (semanticdbEnabled := true)
+          } ++ Seq(
+          semanticdbEnabled := true,
+          // support sbt 1.3.[0-3] which does not contain
+          // https://github.com/sbt/sbt/pull/5202
+          (semanticdbCompilerPlugin := semanticdbCompilerPlugin.value
+            .withRevision((semanticdbVersion).value))
+        )
       settings <-
         inScope(ThisScope.copy(project = Select(project.ref)))(
           scalacOptionsSettings ++ enableSemanticdbPlugin
--- a/src/main/scala/scalafix/sbt/ScalafixEnable.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixEnable.scala
@@ -2,66 +2,161 @@ package scalafix.sbt
 
 import sbt._
 import sbt.Keys._
+import sbt.VersionNumber.SemVer
+
+import ScalafixPlugin.autoImport.scalafixResolvers
+
+import collection.JavaConverters._
+import scala.util._
+import coursierapi.Repository
 
 /** Command to automatically enable semanticdb compiler output for shell session
   */
 object ScalafixEnable {
 
-  /** If the provided Scala binary version is supported, return the full version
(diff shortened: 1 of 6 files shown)


src/sbt-test/sbt-scalafix/dependency/build.sbt
```diff
@@ -1,4 +1,4 @@
-scalaVersion := "2.12.7"
```
bjaglin, line 1: Trying to keep the scalaVersion can cause errors as old semanticdb-scalac are used, for example on java11:
```
sbt:sbt-scalafix> scripted sbt-scalafix/dependency
[info] [info] Using semanticdb-scalac 4.0.0 in project dependency since the version 4.4.32 tracked by scalafix 0.9.34 is not available for scala 2.12.7 - consider upgrading sbt-scalafix or bumping scala
[info] [error] ## Exception when compiling 1 sources to /tmp/sbt_fab36d90/dependency/target/scala-2.12/classes
[info] [error] java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
[info] [error] scala.meta.internal.semanticdb.scalac.InputOps$XtensionGSourceFileInput.toMD5(InputOps.scala:41)
[info] [error] scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument.toTextDocument(TextDocumentOps.scala:614)
[info] [error] scala.meta.internal.semanticdb.scalac.SemanticdbPipeline$SemanticdbTyperComponent$ComputeSemanticdbPhase.saveSemanticdbForCompilationUnit(SemanticdbPipeline.scala:59)
[info] [error] scala.meta.internal.semanticdb.scalac.SemanticdbPipeline$SemanticdbTyperComponent$ComputeSemanticdbPhase.apply(SemanticdbPipeline.scala:65)
[info] [error] scala.tools.nsc.Global$GlobalPhase.applyPhase(Global.scala:441)
[info] [error] scala.tools.nsc.Global$GlobalPhase.run(Global.scala:392)
[info] [error] scala.meta.internal.semanticdb.scalac.SemanticdbPipeline$SemanticdbTyperComponent$ComputeSemanticdbPhase.run(SemanticdbPipeline.scala:75)
[info] [error] scala.tools.nsc.Global$Run.compileUnitsInternal(Global.scala:1467)
```

bjaglin:
cc @olafurpg

adpi2:
I like this effort to make sbt-scalafix as resilient as possible but I would challenge this:

> 2. If not, try to find the earliest binany-compatible semanticdb-scalac (preferred over the latest last to be more repeatable) and use it instead.

There are two cases:
1. we use a recent `scalaVersion` with an old `sbt-scalafix`: In that case the earliest version of `semanticdb-scalac` is better because it is the closest to `sbt-scalafix` so it has greater chances to work. It is repeatable.
2. we use an old  `scalaVersion` with a recent `sbt-scalafix`: in that case however the latest version of `semanticdb-scalac` is probably better because it is the closest to `sbt-scalafix`. It is also repeatable because there should not be any new release of `semanticdb-scala` for that outdated `scalaVersion`.

Also we can adapt the hint correspondingly with either `"consider upgrading sbt-scalafix"` or `"consider bumping scala"`

bjaglin:
> There are two cases

Good catch, I did think about checking the case (mostly for the log message, not the earliest/latest policy) but was not sure it was worth it. Now that I think about it, it would probably solve https://github.com/scalacenter/sbt-scalafix/pull/292#discussion_r833858500. I will add this in a separate commit.

src/sbt-test/sbt-scalafix/dependency/build.sbt
bjaglin: This kind of regression should be mostly addressed by https://github.com/scalacenter/sbt-scalafix/pull/292/commits/b9e14bd32ee0fe18118f8bdf24b7204a77b25f60

src/sbt-test/sbt-scalafix/scalafixEnable/build.sbt
```diff
@@ -15,7 +15,7 @@ lazy val scala211_old = project.settings(
 
 // 2.11.x is supported
 lazy val scala211 = project.settings(
-  // semanticdb-scalac_2.11.11 no longer available, became available as of 2.0.0
```
bjaglin: The fact that there the earliest artifact was non binary-compatible exercised more code, so the change of behavior reduces code coverage unfortunately. Extracting the logic and stubbing the coursier API seems like a lot of work though, so I think we can live with that, and revisit if/when scalameta reaches 5.x.

bjaglin:
The scripted now results in
```
[info] [warn] Forcing scalaVersion 2.11.12 in project scala211_old since no semanticdb-scalac version binary-compatible with 4.4.10 and cross-published for scala 2.11.0 was found - consider bumping scala
[info] [info] Using semanticdb-scalac 4.4.28 in project scala212 since the version 4.4.10 tracked by scalafix 0.9.27 will not be published for scala 2.12.15 - consider upgrading sbt-scalafix
[info] [info] Using semanticdb-scalac 4.1.9 in project scala211 since the version 4.4.10 tracked by scalafix 0.9.27 is no longer published for scala 2.11.11 - consider bumping scala
```
Is that what you had in mind @adpi2 ?

src/main/scala/scalafix/sbt/ScalafixEnable.scala
```diff
@@ -2,66 +2,184 @@ package scalafix.sbt
...
+              coursierapi.Dependency
+                .parse(
+                  project.semanticdbCompilerPlugin0.asCoursierCoordinates,
+                  coursierapi.ScalaVersion.of(project.scalaVersion0.toString)
+                )
+                .getModule
+            val recommendedSemanticdbV =
+              VersionNumber(BuildInfo.scalametaVersion)
+            val compatibleSemanticdbVs = Try(
+              coursierapi.Versions.create
```
bjaglin, line 98: To my knowledge, this is the best Java coursier API to achieve something like suggested in https://github.com/scalacenter/scalafix/issues/1146#issuecomment-1039971557

src/main/scala/scalafix/sbt/ScalafixEnable.scala
```diff
@@ -2,66 +2,184 @@ package scalafix.sbt
...
+                .withRepositories(project.scalafixResolvers0: _*)
```
bjaglin, line 99: That part is untested, so there is a risk that this fails on corporate environments. The failure fallback keeps the current behavior though, so apart from an added latency it shouldn't be a problem.

src/main/scala/scalafix/sbt/ScalafixEnable.scala
```diff
@@ -2,66 +2,184 @@ package scalafix.sbt
...
-  } yield p -> res
+  /** Collect compatible projects across the entire build */
+  private def collectProjects(extracted: Extracted): Seq[CompatibleProject] =
+    for {
+      p <- extracted.structure.allProjectRefs
+      scalaV <- (p / scalaVersion).get(extracted.structure.data).toList
+      partialVersion <- CrossVersion.partialVersion(scalaV).toList
+      maybeRecommendedSemanticdbScalacV <-
+        maybeRecommendedSemanticdbScalacScalaVersion.lift(partialVersion).toList
+      scalafixResolvers0 <- (p / scalafixResolvers)
```
bjaglin, line 45: this is normally set at the `ThisBuild` level, but since usage of custom resolvers is not really tested, I chose to be overly defensive

src/main/scala/scalafix/sbt/ScalafixEnable.scala
adpi2: Maybe this is out of topic in this PR but we could use the `csrConfiguration` key in sbt to configure coursier with authentication and credentials.

bjaglin: I looked at that in the context of https://github.com/scalacenter/scalafix/issues/1571, but I am not sure it would help as [scalafix-core uses the Java coursier API directly](https://github.com/scalacenter/scalafix/pull/1152), not via what lm/lm-coursier.

adpi2: That's true, but it may not be so hard to translate the sbt `csrConfiguration` into the Java API with something similar to [`ToCoursier`](https://github.com/coursier/sbt-coursier/blob/master/modules/lm-coursier/src/main/scala/lmcoursier/definitions/ToCoursier.scala) given that both `csrConfiguration` and the coursier Java API are stable.

bjaglin:
A last tweak on wording and I think we are good to go:
```
[info] Setting semanticdbVersion to 4.4.28 in project scala212 since the version 4.4.10 tracked by scalafix 0.9.27 will not be published for scala 2.12.15 - consider upgrading sbt-scalafix
[info] Setting semanticdbVersion to 4.1.9 in project scala211 since the version 4.4.10 tracked by scalafix 0.9.27 is no longer published for scala 2.11.11 - consider bumping scala
[warn] Forcing scalaVersion to 2.11.12 in project scala211_old since no semanticdb-scalac version binary-compatible with 4.4.10 and cross-published for scala 2.11.0 was found - consider bumping scala
```

adpi2 (approved):
LGTM

PR #324, opened 2022-11-21
avoid deprecated old shell syntax

xuwei-k:
- https://github.com/sbt/sbt/pull/6309

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -32,7 +32,7 @@ object ScalafixPlugin extends AutoPlugin {
       inputKey[Unit](
         "Run scalafix rule(s) in this project and configuration. " +
           "For example: scalafix RemoveUnused. " +
-          "To run on test sources use test:scalafix or scalafixAll. " +
+          "To run on test sources use `Test / scalafix` or scalafixAll. " +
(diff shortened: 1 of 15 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -30,7 +30,7 @@ object ScalafixPlugin extends AutoPlugin {
       inputKey[Unit](
         "Run scalafix rule(s) in this project and configuration. " +
           "For example: scalafix RemoveUnused. " +
-          "To run on test sources use test:scalafix or scalafixAll. " +
+          "To run on test sources use Test/scalafix or scalafixAll. " +
```
bjaglin, line 33: Interestingly, I have always used spaces around `/` - I wonder if there is any reason to?

src/sbt-test/sbt-scalafix/cross-build/test
```diff
@@ -4,18 +4,18 @@
 -> scalafixAll --test ProcedureSyntax
 -> scalafix --test ProcedureSyntax
 -> compile:scalafix --test ProcedureSyntax
--> test:scalafix --test ProcedureSyntax
+-> Test/scalafix --test ProcedureSyntax
```
bjaglin, line 7: As is, this PR seems to only handle shell commands on the Test config, so we would continue to have warnings (and eventually errors). Maybe the regexp you use could be updated?

bjaglin (commented):
Thanks for initiating that!

PR #348, opened 2023-06-03
Scalafix 0.11.0

bjaglin:


src/test/scala/scalafix/internal/sbt/ScalafixAPISuite.scala
```diff
@@ -11,6 +11,7 @@ import org.scalatest.funsuite.AnyFunSuite
 
 import scala.util.Properties
 
+@org.scalatest.Ignore
```
bjaglin, line 14: https://github.com/scalacenter/example-scalafix-rule/issues/11

PR #351, opened 2023-06-11
expose low-level API for scalafix repo dogfooding

bjaglin:


bjaglin:
as we ended up having settings depending on tasks, discontinued in favor of using a command to mutate the `plugin` project's `libraryDependencies`

PR #378, opened 2023-11-12
improve --files CLI UX

bjaglin:
Fixes https://github.com/scalacenter/scalafix/issues/1874

To avoid getting false positive errors (missing SemanticDB in presence of semantic rules), scalafix executions with `--files` should not aggregate as they are by nature specific to a given project / Configuration.

- prevent usage of `scalafixAll --files`
- validate `--files` values against `unmanagedSourceDirectories`
  - allow precise completions
  - avoid aggregated executions: since requested files are not valid outside the project on which the task input was crafted, sbt simply ignores the runs


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -127,13 +127,6 @@ object ScalafixPlugin extends AutoPlugin {
       Invisible
     )
 
-  private val scalafixCompletions: SettingKey[ScalafixCompletions] =
```
bjaglin, line 130: inlining this saves some setting keys (sbt-scalafix adds a lot...)

src/sbt-test/sbt-scalafix/explicit-files/.scalafix.conf
```diff
@@ -1,2 +1,2 @@
-rules = [DisableSyntax]
+rules = [DisableSyntax, RemoveUnused]
```
bjaglin, line 1: this is needed to trigger the "semanticdb not found" error from the original bug report

src/sbt-test/sbt-scalafix/explicit-files/test
```diff
@@ -1,14 +1,38 @@
...
\ No newline at end of file
+# Targeting invalid files fails
+-> ok_and_ko / scalafix -f=ok_and_ko/src/main/scala/KO.scala
+
+# Targeting valid and invalid files fails
+-> ok_and_ko / scalafix -f=ok_and_ko/src/main/scala/OK.scala -f=ok_and_ko/src/main/scala/KO.scala
+-> ok_and_ko / scalafix -f=ok_and_ko/src/main/scala/KO.scala -f=ok_and_ko/src/main/scala/OK.scala
+
+# Targeting valid files from the root project succeeds
+> scalafix -f=src/main/scala/OK.scala
```
bjaglin, line 22: implemented in this PR

src/sbt-test/sbt-scalafix/explicit-files/test
```diff
@@ -1,14 +1,38 @@
...
+-> ok_only / scalafix -f=ok_only/src/main/scala/NotHere.scala
+
+# Targeting with scalafixAll valid files from the correct project fails
+-> ok_only / scalafixAll -f=ok_only/src/main/scala/OK.scala
+
+# Targeting valid files from a different project fails
+-> ok_only / scalafix -f=ok_and_ko/src/main/scala/OK.scala
+
+# Targeting valid files from a different configuration fails
+-> ok_only / Test / scalafix -f=ok_only/src/main/scala/OK.scala
```
bjaglin, line 38: implemented in this PR

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -343,12 +337,23 @@ object ScalafixPlugin extends AutoPlugin {
...
       config: Configuration
-  ): Def.Initialize[InputTask[Unit]] =
+  ): Def.Initialize[InputTask[Unit]] = {
+    val parser = Def.setting(
+      new ScalafixCompletions(
+        workingDirectory = (ThisBuild / baseDirectory).value.toPath,
+        loadedRules = () => scalafixInterfaceProvider.value().availableRules(),
+        terminalWidth = Some(JLineAccess.terminalWidth),
+        allowedTargetFilesPrefixes =
+          (scalafix / unmanagedSourceDirectories).value.map(_.toPath)
```
bjaglin, line 347: we don't rely on `unmanagedSources` here as
1. we want to be able to target directories as well
2. it's a task, not available at settings evaluation time anyway

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: This caused a regression, see https://github.com/scalacenter/sbt-scalafix/pull/405

PR #380, opened 2023-11-30
Support several scalafixScalaBinaryVersion within a build

zarthross:
Because the `scalafixInterfaceProvider` was getting its scala versions from the `ThisBuild` level, scalafix would not use a different binary version for each scala version when running with `++scalafix`.  This limited users ability to use scala 2.13 only rules AND use scalafix on scala 3.

This change would make it possible to run the appropriate scalafix binary as the scala version changes during cross building (because `ThisBuild / scalaVersion` does not change during cross building), making it possible to dynamically update the scalafix.conf file for each scala version.  In the future a sbt key could be added to point at different scalafix.conf files for different scala versions if that functionality is desired.


zarthross:
@armanbilge So, that really isn't sufficient.  While that would change the scalafix file based on the scala version, the problem is the scalafix binary version

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -158,6 +158,11 @@ object ScalafixPlugin extends AutoPlugin {
         SettingKey[Boolean]("bspEnabled") := false
       )
     ),
+    scalafixInterfaceProvider := ScalafixInterface.fromToolClasspath(
```
bjaglin, line 161: I didn't check but this will likely result in a significant memory increase as a separate instance of scalafix (together with the scala standard lib) will be classloaded for **each** project. Memoizing the provider per scalafixScalaBinaryVersion should do the trick, but I guess we will need to add some custom code as we can't rely on sbt scope to do that.

src/sbt-test/sbt-scalafix/cross-build-advanced/project/plugins.sbt
```diff
@@ -0,0 +1,3 @@
+resolvers += Resolver.sonatypeRepo("public")
+libraryDependencies += "com.googlecode.java-diff-utils" % "diffutils" % "1.3.0"
```
bjaglin, line 2: probably unused?

bjaglin (commented):
Thanks for the PR! As a follow up to the comment, I would suggest renaming it to "Support several scalafixScalaBinaryVersion within a build" to better carry the intent; being able to have a different configuration file to disable ExplicitResultTypes on scala3 is a related feature, which is already available.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
zarthross: If I'm reading the code right, there might be duplication in the event there isn't any custom tool paths, but we can correct for that by change the cache to be keyed on an `Option[ToolClasspath]` instead of just a `ToolClasspath` and we can create the default once and store it there.

zarthross: @bjaglin Would you like me to make that cache change?  Is there some other change I can make to get this accepted?

bjaglin: > If I'm reading the code right, there might be duplication in the event there isn't any custom tool paths, but we can correct for that by change the cache to be keyed on an Option[ToolClasspath] instead of just a ToolClasspath and we can create the default once and store it there.

My concern is that `ScalafixInterface.fromToolClasspath()` is now called once per project instead of once per build, and there is no memoization inside. The fact that the key was at the build level guaranteed the singleton nature of the provider. The `ToolClasspath` [layering](https://github.com/bjaglin/sbt-scalafix/blob/568f1a2e435b225117aacae59391d505da9f1b6a/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala#L15-L31) is handled separately and won't be directly affected by this change.

> Would you like me to make that cache change?

Yes, I am afraid I can't get this as-is as it brings a serious regression. Although I am not sure we were talking about the same change, see my clarification above.

To illustrate my previous comment, I monitored the RAM usage of `scalafixAll --no-cache` on a fresh sbt session, against a warm build with many projects (scalafix). Note that this build uses sbt-projectmatrix and [only run scalafix on scala 2.13 projects](https://github.com/scalacenter/scalafix/blob/11a95ff17954518b0f28b0798654f7b1af0d4cd2/project/ScalafixBuild.scala#L306-L319) (for the same reason you are trying to address), but that's irrelevant here. I forced a GC a few seconds before I took the screenshot in both cases.

### sbt-scalafix 0.11.1+23-b71ae18a-SNAPSHOT (this PR)

Screenshot from 2023-12-04 23-06-10

```
[success] Total time: 72 s (01:12), completed Dec 4, 2023, 11:01:13 PM
```

### sbt-scalafix 0.11.1

Screenshot from 2023-12-04 23-10-17

```
[success] Total time: 54 s, completed Dec 4, 2023, 11:08:16 PM
```

zarthross: @bjaglin Pushed an update that memoizes `fromToolClasspath`, let me know if this works for you.

src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
```diff
@@ -116,47 +116,60 @@ object ScalafixInterface {
...
-        )
-      else if (scalafixBinaryScalaVersion == "2.11")
-        logger.error(
-          "Scala 2.11 is no longer supported. Please downgrade to the final version supporting " +
-            "it: sbt-scalafix 0.10.4."
-        )
-      val callback = new ScalafixLogger(logger)
-      val scalafixArguments = ScalafixAPI
-        .fetchAndClassloadInstance(
+      fromToolClasspathMemo.getOrElseUpdate(
```
bjaglin, line 131: As this will be racy and the cost of loading stuff below is high (involving IO), I believe we should favor a Map implementation with blocking methods over one with only pure CAS methods. Unless I am mistaken, `ConcurrentHashMap.computeIfAbsent` does provide the guarantee that the block is never called more than once per key, while nothing in the Scala collections does. Could you update the implementation? Thanks!

src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
```diff
@@ -116,47 +116,60 @@ object ScalafixInterface {
     private lazy val _value = scala.util.Try(thunk())
     override def apply(): T = _value.get
   }
+  private val fromToolClasspathMemo: scala.collection.mutable.HashMap[
```
bjaglin, line 119: This instance will persist after a `sbt reload`, so we might accumulate stale `ScalafixInterface` instances if the user keeps changing the build definition in a way that impacts the composite key below. The alternative might be to keep that cache in a build-level setting key, but I am not even sure the classloader would reclaim everything when the old cache/setting value gets garbage-collected after a `sbt reload`? :thinking: 

Let's keep this as-is considering the cardinality is very low :+1:

src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
zarthross: Updated to the existing `scalafix.internal.sbt.BlockingCache` which uses a `java.util.Collections.synchronizedMap` which should prevent race conditions.  Since the cardinality is low and the key access is only on scalafix on a per-project basis, this should also not block very often at all.

bjaglin: Perfect, I had forgotten about that one!

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -158,6 +158,11 @@ object ScalafixPlugin extends AutoPlugin {
...
+      (scalafixScalaBinaryVersion).value,
+      scalafixDependencies = (scalafixDependencies).value,
+      scalafixCustomResolvers = (scalafixResolvers).value
```
bjaglin, line 164: nit: unnecessary parens
```suggestion
      scalafixScalaBinaryVersion.value,
      scalafixDependencies = scalafixDependencies.value,
      scalafixCustomResolvers = scalafixResolvers.value
```

src/sbt-test/sbt-scalafix/cross-build-advanced/build.sbt
```diff
@@ -0,0 +1,24 @@
...
+lazy val root = project
+  .in(file("."))
+  .settings(
+    scalaVersion := Versions.scala212,
+    crossScalaVersions := Seq(Versions.scala212, Versions.scala213, scala3Version),
+    scalacOptions ++= (if (scalaVersion.value.startsWith("2")) Seq("-Ywarn-unused") else Seq()),
+    scalafixScalaBinaryVersion := {
+      if ((scalaBinaryVersion).value == "3") {
+        (scalafixScalaBinaryVersion).value
+      } else { (scalaBinaryVersion).value }
```
bjaglin, line 16: nit: simplify syntax
```suggestion
      if (scalaBinaryVersion.value == "3") scalafixScalaBinaryVersion.value
      else scalaBinaryVersion.value
```

src/sbt-test/sbt-scalafix/cross-build-advanced/.scalafix-2.conf
```diff
@@ -0,0 +1,3 @@
+rules = [DisableSyntax, OrganizeImports, ExplicitResultTypes]
```
bjaglin, line 1: DisableSyntax seems unexercised, but it doesn't hurt to keep a syntactic rule here I guess

bjaglin (approved):
Thanks a lot @zarthross, this will be a great addition! I'll open a PR shortly to update the docs.

For the record, I am planning to do a release some time next month, in sync with the [scala 2.x patch releases](https://contributors.scala-lang.org/t/scala-2-12-19-release-planning/6216/3). Until then, you can try the [snapshots](https://github.com/scalacenter/sbt-scalafix/tree/main#nightlies).

src/sbt-test/sbt-scalafix/cross-build-advanced/build.sbt
```diff
@@ -0,0 +1,23 @@
+import _root_.scalafix.sbt.{BuildInfo => Versions}
+
+val scala3Version = "3.3.0"
+ThisBuild / semanticdbEnabled := true
+ThisBuild / semanticdbVersion := scalafixSemanticdb.revision
```
bjaglin, line 5:
```suggestion
Global / semanticdbVersion := scalafixSemanticdb.revision // support sbt 1.3.x, see https://github.com/sbt/sbt/issues/5110
```

bjaglin:
https://github.com/scalacenter/sbt-scalafix/pull/380/commits/127137096f5afc6b76c526207bcd14e8e887075d had more impact than I thought (which is good to expose anyway). https://github.com/scalacenter/sbt-scalafix/pull/380/commits/c96467459dcd95365dc1df6a411f26f7f9bd4b9b should get that back to green before merging.

PR #384, opened 2024-01-23
add `scalafixCallback` setting key to customize diagnostics reporting

xuwei-k:
Towards https://github.com/scalacenter/scalafix/issues/951

### DIFF
--- a/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
+++ b/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
@@ -127,20 +127,6 @@ object ScalafixInterface {
       scalafixBinaryScalaVersion: String,
       scalafixDependencies: Seq[ModuleID],
       scalafixCustomResolvers: Seq[Repository],
-      logger: Logger = defaultLogger
-  ): () => ScalafixInterface =
-    fromToolClasspath(
-      scalafixBinaryScalaVersion = scalafixBinaryScalaVersion,
-      scalafixDependencies = scalafixDependencies,
-      scalafixCustomResolvers = scalafixCustomResolvers,
-      logger = logger,
-      callback = new ScalafixLogger(logger)
-    )
-
-  def fromToolClasspath(
-      scalafixBinaryScalaVersion: String,
-      scalafixDependencies: Seq[ModuleID],
-      scalafixCustomResolvers: Seq[Repository],
       logger: Logger,
       callback: ScalafixMainCallback
   ): () => ScalafixInterface =
--- a/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
+++ b/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
@@ -121,11 +121,28 @@ object ScalafixInterface {
     ScalafixInterface
   ] = new BlockingCache
 
+  private[scalafix] val defaultLogger: Logger = ConsoleLogger(System.out)
+
   def fromToolClasspath(
       scalafixBinaryScalaVersion: String,
       scalafixDependencies: Seq[ModuleID],
       scalafixCustomResolvers: Seq[Repository],
-      logger: Logger = ConsoleLogger(System.out)
+      logger: Logger = defaultLogger
+  ): () => ScalafixInterface =
+    fromToolClasspath(
+      scalafixBinaryScalaVersion = scalafixBinaryScalaVersion,
+      scalafixDependencies = scalafixDependencies,
+      scalafixCustomResolvers = scalafixCustomResolvers,
+      logger = logger,
+      callback = new ScalafixLogger(logger)
+    )
+
+  def fromToolClasspath(
+      scalafixBinaryScalaVersion: String,
+      scalafixDependencies: Seq[ModuleID],
+      scalafixCustomResolvers: Seq[Repository],
+      logger: Logger,
+      callback: ScalafixMainCallback
   ): () => ScalafixInterface =
     new LazyValue({ () =>
       fromToolClasspathMemo.getOrElseUpdate(
@@ -145,7 +162,6 @@ object ScalafixInterface {
               "Scala 2.11 is no longer supported. Please downgrade to the final version supporting " +
                 "it: sbt-scalafix 0.10.4."
             )
-          val callback = new ScalafixLogger(logger)
           val scalafixArguments = ScalafixAPI
             .fetchAndClassloadInstance(
               scalafixBinaryScalaVersion,
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -9,7 +9,7 @@ import sbt.Keys._
 import sbt._
 import sbt.internal.sbtscalafix.JLineAccess
 import sbt.plugins.JvmPlugin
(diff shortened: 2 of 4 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -158,10 +162,13 @@ object ScalafixPlugin extends AutoPlugin {
         SettingKey[Boolean]("bspEnabled") := false
       )
     ),
+    scalafixCallback := new ScalafixLogger(ScalafixInterface.defaultLogger),
```
bjaglin, line 165: could you move this default up to `globalSettings` to save some RAM?

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -75,6 +75,10 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional location to .scalafix.conf file to specify which scalafix rules should run. " +
           "Defaults to the build base directory if a .scalafix.conf file exists."
       )
+
+    val scalafixCallback: SettingKey[ScalafixMainCallback] =
+      settingKey[ScalafixMainCallback]("")
```
bjaglin, line 80:
```suggestion
      settingKey[ScalafixMainCallback](
        "The handler for the diagnostics emitted during scalafix execution. " +
          "Defaults to a wrapper around `sbt.Logger`."
      )
```

src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
```diff
@@ -121,11 +121,28 @@ object ScalafixInterface {
...
+      scalafixCustomResolvers: Seq[Repository],
+      logger: Logger = defaultLogger
+  ): () => ScalafixInterface =
+    fromToolClasspath(
+      scalafixBinaryScalaVersion = scalafixBinaryScalaVersion,
+      scalafixDependencies = scalafixDependencies,
+      scalafixCustomResolvers = scalafixCustomResolvers,
+      logger = logger,
+      callback = new ScalafixLogger(logger)
+    )
```
bjaglin, line 138: This class is in an internal package so no need to maintain backward compatibility. A test or two might need to be updated though.
```suggestion
```

bjaglin (commented):
Thanks for the PR. I am curious about your use-case. I assume it's similar to what was discussed in https://github.com/scalacenter/scalafix/issues/951?

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: For the record, your PR made me realize docs were not up to date: https://github.com/scalacenter/sbt-scalafix/pull/385.

bjaglin (commented):
As I read again my "Can be set in ThisBuild or at project-level" suggestion, I realized I missed an important point in my first review - see inline comment below.

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -158,10 +162,13 @@ object ScalafixPlugin extends AutoPlugin {
...
       )
     ),
+    scalafixCallback := new ScalafixLogger(ScalafixInterface.defaultLogger),
     scalafixInterfaceProvider := ScalafixInterface.fromToolClasspath(
       scalafixScalaBinaryVersion.value,
       scalafixDependencies = scalafixDependencies.value,
-      scalafixCustomResolvers = scalafixResolvers.value
+      scalafixCustomResolvers = scalafixResolvers.value,
+      logger = ScalafixInterface.defaultLogger,
+      callback = scalafixCallback.value
```
bjaglin, line 171: Watch out for:
https://github.com/scalacenter/sbt-scalafix/blob/c2777cfe1ea5623201fc3ae86d74e1043e19f4f3/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala#L131-L136

As it stands in the current state of this PR, a client might set different values for `scalafixCallback` across projects, and see only one being used because of memoization.

We can either
1. add `callback` as a cache key (using reference equality I guess?)
1. support only one value per build
```suggestion
      callback = (ThisBuild / scalafixCallback).value
```

I would go for (1) for consistency with other configurable attributes of `scalafixInterface`

xuwei-k:
> I assume it's similar to what was discussed in https://github.com/scalacenter/scalafix/issues/951?

Oh, I didn't know the issue. I have same motivation 👍

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -75,6 +75,13 @@ object ScalafixPlugin extends AutoPlugin {
         "Optional location to .scalafix.conf file to specify which scalafix rules should run. " +
           "Defaults to the build base directory if a .scalafix.conf file exists."
       )
+
+    val scalafixCallback: SettingKey[ScalafixMainCallback] =
+      settingKey[ScalafixMainCallback](
+        "The handler for the diagnostics emitted during scalafix execution. Can be set in ThisBuild or at project-level. " +
```
bjaglin, line 81: Since you went for (2) in https://github.com/scalacenter/sbt-scalafix/pull/384#discussion_r1462919626
```suggestion
        "The handler for the diagnostics emitted during scalafix execution. Must be set in ThisBuild. " +
```

bjaglin (approved):

bjaglin (approved):

bjaglin:
> > I assume it's similar to what was discussed in [scalacenter/scalafix#951](https://github.com/scalacenter/scalafix/issues/951)?
> 
> Oh, I didn't know the issue. I have same motivation 👍

Nice. I just looked at your projects and I assume this PR is effectively a step towards https://github.com/xuwei-k/sbt-warning-diff/issues/6. I'll be watching that - if you manage to do what you want with a custom `ScalafixMainCallback`, we could upstream it to the main scalafix repo, maybe in a new `contrib` module and close https://github.com/scalacenter/scalafix/issues/951. Let' see!

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin: Reopening this for visibility: given the current state of sbt-scalafix, there is a small limitation to be aware of in the solution (2) you went for. The [cache is never flushed](https://github.com/scalacenter/sbt-scalafix/blob/a540aebe45b62f67530d11b484f33b3a46624784/src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala#L119-L122), so if a sbt command changed the state of the build and update `ThisBuild / scalafixCallback`, you will get false positive cache hits.

This is a corner case, but depending on how you end up injecting a per-invocation `scalafixCallback` to avoid mutability at that level, this could be annoying. If that's the case, we'll have to go for (1), as well as maybe keep the cache in a setting key to avoid leaking memory upon each `sbt reload` because of a new JVM instance of `ScalafixMainCallback`.

bjaglin:
For the record, I am planning to do a Scalafix release in a few weeks, in sync with the [scala 2.x patch releases](https://contributors.scala-lang.org/t/scala-2-12-19-release-planning/6216/3). Until then, you can try the [snapshots](https://github.com/scalacenter/sbt-scalafix/tree/main#nightlies).

PR #405, opened 2024-03-23
add `scalafixJGitCompletion` key. reduce memory usage

xuwei-k:
Maybe partially(?) fix https://github.com/scalacenter/scalafix/issues/1965 if there are many sub projects

if there is example sbt project as follow

```scala
lazy val p1 = project
lazy val p2 = project
lazy val p3 = project
lazy val p4 = project.disablePlugins(ScalafixPlugin)
```

### before

(`p1` + `p2` + `p3` + `implicit root project`) * ( `Compile/scalafix` + `Test/scalafix` + `scalafixAll` )
= 4 * 3
= 12`JGitCompletion` instances

### after

only one `JGitCompletion` instance

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -118,6 +118,13 @@ object ScalafixPlugin extends AutoPlugin {
 
   import autoImport._
 
+  private val scalafixJGitCompletion: SettingKey[JGitCompletion] =
+    SettingKey(
+      "scalafixJGitCompletion",
+      "Implementation detail - do not use",
+      Invisible
+    )
+
   private val scalafixDummyTask: TaskKey[Unit] =
     TaskKey(
       "scalafixDummyTask",
@@ -233,7 +240,10 @@ object ScalafixPlugin extends AutoPlugin {
 
   override def buildSettings: Seq[Def.Setting[_]] =
     Seq(
-      scalafixScalaBinaryVersion := "2.12"
+      scalafixScalaBinaryVersion := "2.12",
+      scalafixJGitCompletion := new JGitCompletion(
+        (ThisBuild / baseDirectory).value.toPath
+      )
     )
 
   lazy val stdoutLogger = ConsoleLogger(System.out)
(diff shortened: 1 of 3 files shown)


src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -299,7 +309,8 @@ object ScalafixPlugin extends AutoPlugin {
         workingDirectory = (ThisBuild / baseDirectory).value.toPath,
         loadedRules = () => scalafixInterfaceProvider.value().availableRules(),
         terminalWidth = Some(JLineAccess.terminalWidth),
-        allowedTargetFilesPrefixes = Nil
+        allowedTargetFilesPrefixes = Nil,
+        jgitCompletion = (ThisBuild / scalafixJGitCompletion).value
```
bjaglin, line 313: This shouldn't matter for most cases, but for the sake of potential extensibility we could let potential project-level overrides
```suggestion
        jgitCompletion = scalafixJGitCompletion.value
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -352,7 +363,8 @@ object ScalafixPlugin extends AutoPlugin {
         loadedRules = () => scalafixInterfaceProvider.value().availableRules(),
         terminalWidth = Some(JLineAccess.terminalWidth),
         allowedTargetFilesPrefixes =
-          (scalafix / unmanagedSourceDirectories).value.map(_.toPath)
+          (scalafix / unmanagedSourceDirectories).value.map(_.toPath),
+        jgitCompletion = (ThisBuild / scalafixJGitCompletion).value
```
bjaglin, line 367: This shouldn't matter for most cases, but for the sake of potential extensibility we could let potential project-level overrides
```suggestion
        jgitCompletion = scalafixJGitCompletion.value
```

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -233,7 +240,10 @@ object ScalafixPlugin extends AutoPlugin {
 
   override def buildSettings: Seq[Def.Setting[_]] =
     Seq(
-      scalafixScalaBinaryVersion := "2.12"
+      scalafixScalaBinaryVersion := "2.12",
+      scalafixJGitCompletion := new JGitCompletion(
+        (ThisBuild / baseDirectory).value.toPath
+      )
```
bjaglin, line 246: The explicit `ThisBuild` scope looks redundant here
```suggestion
      scalafixJGitCompletion := new JGitCompletion(baseDirectory.value.toPath)
```

bjaglin (commented):
Thanks for looking into that @xuwei-k !

PR #408, opened 2024-04-15
fix JGitCompletions memory leak

xuwei-k:
- https://github.com/scalacenter/scalafix/issues/1965#issuecomment-2016445243


bjaglin (approved):
Thanks you!

PR #412, opened 2024-04-22
drop sbt 1.3.x support & speedup CI by parallelizing scripted

bjaglin:
### Goals
- simplify the build
- run scripted tests in parallel (20min -> 8min for CI)

### Why can we do it?
- [sbt 1.4.0 is 4 years old](https://github.com/sbt/sbt/releases/v1.4.0)
- there is no good reason to stick to an old sbt minor

### TODO
- [x] [docs updated](https://github.com/scalacenter/scalafix/pull/1980)

PR #413, opened 2024-04-23
parallelize scripted

bjaglin:
Based on https://github.com/scalacenter/sbt-scalafix/pull/412

Blocker: batch mode effectively ignores the custom `build.properties`


build.sbt
```diff
@@ -73,6 +73,8 @@ scalacOptions ++= List(
 enablePlugins(ScriptedPlugin)
 sbtPlugin := true
 scriptedBufferLog := false
+scriptedBatchExecution := true
```
bjaglin, line 76: despite the intent of https://github.com/sbt/sbt/pull/6151, `scriptedBatchExecution` is never true by default (because [`binVersion`](https://github.com/sbt/sbt/pull/6151/files#diff-6b2c28f765181fcabca3279c1f686a4ca39ece8ad09b2ec4281dc9287dc4fdc5R84) is "1" on sbt 1.x)

bjaglin:
Reintegrating into https://github.com/scalacenter/sbt-scalafix/pull/412

PR #415, opened 2024-04-24
reduce caching scripted test runtime by limiting number of reloads

bjaglin:


bjaglin:
there is no significant change, the test still takes ~2min

PR #419, opened 2024-05-01
let scalafixScalaBV follow scalaBV to leverage 2.13 & deprecate it

bjaglin:
## Rationale

* Fixes https://github.com/scalacenter/scalafix/issues/1592 as demonstrated in https://github.com/scalacenter/scalafix/pull/1991.
* Most of the Scalafix users that do not use `ExplicitResultTypes` are running Scalafix with Scala 2.12 (since they don't need to set a custom `scalafixScalaBinaryVersion`). However, Scala 2.13 collections are known to perform better (and if they don't, we need to fix rules since that's the baseline for the future as Scala 3 uses them anyway). [Concerns raised in the past](https://github.com/scala/bug/issues/11627#issuecomment-514619316) are no longer relevant in recent versions of [OpenJDK](https://bugs.openjdk.org/browse/JDK-8234863) or GraalVM. As a result, we want to push for using Scala 2.13 whenever possible

## Implementation

Instead of moving to 2.13 on all builds that do not have `scalafixScalaBinaryVersion` explicitly set (which would be a breaking change for users of `ExplicitResultTypes` on 2.12 sources), the default now honors the Scala binary version at the project level (leveraging https://github.com/scalacenter/sbt-scalafix/pull/380), falling back to Scala 2.13 for Scala 3 projects until https://github.com/scalacenter/scalafix/issues/1680 is sorted out.

### Published community rules

We have been advocating for cross-publishing rules for 4 years (for [new](https://github.com/scalacenter/scalafix/issues/1202) and [existing](https://github.com/scalacenter/scalafix/pull/1237) projects), so all popular rules are now cross-published. As a result, running community rules with Scalafix in Scala 2.13 on projects with Scala 2.13 or Scala 3 should not cause artifact resolution issues.

### Unpublished/source community rules

Running unpublished rules relies on on-the-fly compilation via scalafix-reflect, using the Scala compiler of the runtime Scala version. Even though some users were customizing the Scala version to use `ExplicitResultTypes` and therefore potentially exercising compilation of some unpublished rules with Scala 2.13, that was the minority.

This changes officializes the need for unpublished rules to support both Scala 2.12 and Scala 2.13 for a maximum reach. The biggest challenge is collection compatibility, but scalafix-core pulls [scala-collection-compat](https://github.com/scala/scala-collection-compat) since [0.9.7](https://github.com/scalacenter/scalafix/pull/980) (released 4,5 years ago), so rule authors have been able to rely on the Scala 2.13 collection framework syntax for as long as Scalafix is mature.

It's hard to tell what's the proportion of unpublished rules which can compile with Scala 2.12 but not with Scala 2.13. Users of those will be impacted in the following scenarios:
* sbt-scalafix users with projects with Scala 2.13 or 3 sources, not overriding `scalafixScalaBinaryVersion`
* [Scala Steward migrations](https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/resources/scalafix-migrations.conf) ([as soon as an sbt-scalafix release containing this PR is cut no matter if it's a minor or a patch](https://github.com/scala-steward-org/scala-steward/blob/996af6a8b62b210637ba5fccc37d073f7131f6e2/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala#L129-L132))


src/sbt-test/sbt-1.5/cross-build-scala3/.scalafix-3.conf
```diff
@@ -0,0 +1 @@
+rules = [OrganizeImports]
```
bjaglin, line 1: this is not detected as a rename, because I removed the unexercised `DisableSyntax`

src/sbt-test/sbt-1.5/cross-build-scala3/build.sbt
```diff
@@ -16,10 +16,6 @@ lazy val root = project
     scalacOptions ++= (if (scalaVersion.value.startsWith("2"))
                          Seq("-Ywarn-unused")
                        else Seq()),
-    scalafixScalaBinaryVersion := {
-      if (scalaBinaryVersion.value == "3") scalafixScalaBinaryVersion.value
-      else scalaBinaryVersion.value
-    },
```
bjaglin, line 22: this was a no-op, keeping the default of "2.12"

src/sbt-test/sbt-1.5/cross-build-scala3/src/main/scala-3/Main.scala.expected
```diff
@@ -0,0 +1,5 @@
+import scala.Int
```
bjaglin, line 1: this is not detected as a rename because I enabled removal of unused imports to make sure a rule was run (unrelated to the fix)

src/sbt-test/sbt-1.5/cross-build-scala3/build.sbt
```diff
@@ -13,12 +13,9 @@ lazy val root = project
       Versions.scala213,
       scala3Version
     ),
-    scalacOptions ++= (if (scalaVersion.value.startsWith("2"))
-                         Seq("-Ywarn-unused")
-                       else Seq()),
-    scalafixScalaBinaryVersion := {
-      if (scalaBinaryVersion.value == "3") scalafixScalaBinaryVersion.value
-      else scalaBinaryVersion.value
```
bjaglin, line 21: this was a no-op, keeping the default of "2.12"

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -64,10 +64,15 @@ object ScalafixPlugin extends AutoPlugin {
...
       )
+    @deprecated(
+      "scalafixScalaBinaryVersion now follows scalaBinaryVersion by default",
+      "0.12.1"
+    )
     val scalafixScalaBinaryVersion: SettingKey[String] =
       settingKey[String](
-        "The Scala binary version used for scalafix execution. Can be set in ThisBuild or at project-level. " +
-          "Custom rules must be compiled against that binary version. Defaults to 2.12."
+        "The Scala binary version used for scalafix execution. Must be set at project-level. " +
```
bjaglin, line 73: Usage of `derive()` prevents visibility of user-defined `ThisBuild / scalafixScalaBinaryVersion` (which was the only option until [recently](https://github.com/scalacenter/sbt-scalafix/pull/380)) at project-level. That said, I can't think of a use case where the user would need to force a version which does not match scalaBinaryVersion (thus the deprecation).

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -266,6 +271,13 @@ object ScalafixPlugin extends AutoPlugin {
   )
 
   override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
+    // use derive to support per-level values while allowing user to override at ThisBuild level
```
bjaglin, line 274: the possible build-level override is `scalaVersion`, not `scalafixScalaBinaryVersion` as explained above

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
bjaglin, line 73: Until [recently](https://github.com/scalacenter/sbt-scalafix/pull/380), `ThisBuild` was the only place where `scalafixScalaBinaryVersion` could be set, so most of the clients must be setting it at that level currently. However, that should not be a problem because the only reason for them to set it was to follow `scalaScalafixBinaryVersion`, which is now the default (at project-level, thus the inability to override it at `ThisBuild` level).

That's why the key is being deprecated, and the change in where it effectively not a breaking change.

PR #453, opened 2025-01-11
test current scalafix against 2.13.16 sources

bjaglin:
https://github.com/scalameta/scalameta/issues/4128#issuecomment-2585373710


src/sbt-test/sbt-scalafix/semanticdb-enabled/build.sbt
```diff
@@ -1,5 +1,4 @@
-// support sbt 1.3.x, see https://github.com/sbt/sbt/issues/5110
-Global / semanticdbVersion := scalafixSemanticdb.revision
```
bjaglin, line 2: should have been cleaned up in https://github.com/scalacenter/sbt-scalafix/pull/412, just doing some boy-scout here since I touched the file

PR #456, opened 2025-01-11
make sure sbt-scalafix is published with JDK8 bytecode

bjaglin:
~Fix regression introduced in https://github.com/scalacenter/sbt-scalafix/pull/414 (impacting 0.12.1 & 0.13.0)~

... actually https://github.com/scalacenter/sbt-scalafix/blob/9f32b14f26d952659f8758f5b469e86db16a438d/build.sbt#L66 is enough, but better safe than sorry!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,7 @@ jobs:
       - uses: coursier/setup-action@v1
         with:
           apps: sbt
+          jvm: temurin:8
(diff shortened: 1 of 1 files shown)


bjaglin:
actually https://github.com/scalacenter/sbt-scalafix/blob/9f32b14f26d952659f8758f5b469e86db16a438d/build.sbt#L66 is enough

PR #463, opened 2025-01-31
Test more against Scala 3 projects

bjaglin:


src/sbt-test/skip-sbt1.4/cross-build-scala3/build.sbt
```diff
@@ -1,6 +1,6 @@
 import _root_.scalafix.sbt.{BuildInfo => Versions}
 
-val scala3Version = "3.4.1"
+val scala3Version = "3.3.5"
```
bjaglin, line 3: this requests cli_3.5.x which brings the compile, therefore we need the unreleased https://github.com/scalacenter/scalafix/pull/2183 for JDK23 to work

PR #467, opened 2025-02-13
Enhance error message when SemanticDB scalac plugin can't be found

danicheg:
Hey there! Recently, I've been upgrading some dependencies of one project and accidentally got this error message in SBT:

> The SemanticDB scalac plugin version 4.9.3 set up via `semanticdbVersion` does follow the version recommended for Scalafix, but is not supported for the outdated Scala version 2.13.16. Please upgrade to a more recent Scala patch version or uninstall sbt-scalafix.

Honestly, it's a bit confusing since Scala 2.13.16 is the most recent version of the 2.13 series. What I actually had to do was update the sbt-scalafix version to the latest one.

So, here's the enhancement of that error message.

### DIFF
--- a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -250,11 +250,12 @@ object ScalafixPlugin extends AutoPlugin {
               resolveException.failed.headOption match {
                 case Some(SemanticdbScalac(rev)) =>
                   val scalaV = scalaVersion.value
-                  val msg = s"The SemanticDB scalac plugin version ${rev} set up " +
-                    "via `semanticdbVersion` does follow the version recommended " +
-                    "for Scalafix, but is not supported for the outdated Scala " +
-                    s"version ${scalaV}. Please upgrade to a more recent Scala " +
-                    "patch version or uninstall sbt-scalafix."
+                  val msg =
+                    s"The SemanticDB scalac plugin version ${rev} set up " +
+                      "via `semanticdbVersion` does follow the version recommended " +
+                      "for Scalafix, but is not supported for the given Scala " +
+                      s"version ${scalaV}. Please consider upgrading to a more recent version " +
+                      "of sbt-scalafix and/or Scala, or uninstalling sbt-scalafix plugin."
                   throw inc.copy(message = Some(msg))
                 case _ =>
               }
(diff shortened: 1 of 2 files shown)


danicheg:
cс @bjaglin

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -252,9 +252,10 @@ object ScalafixPlugin extends AutoPlugin {
                   val scalaV = scalaVersion.value
                   val msg = s"The SemanticDB scalac plugin version ${rev} set up " +
                     "via `semanticdbVersion` does follow the version recommended " +
-                    "for Scalafix, but is not supported for the outdated Scala " +
-                    s"version ${scalaV}. Please upgrade to a more recent Scala " +
-                    "patch version or uninstall sbt-scalafix."
+                    "for Scalafix, but is not supported for the given Scala " +
+                    s"version ${scalaV}. Please consider upgrading to a more recent version " +
+                    "of sbt-scalafix and/or Scala, or uninstalling sbt-scalafix plugin. " +
+                    "Crucially, sbt-scalafix must be published for the given Scala version."
```
bjaglin, line 258: Can you clarify why you think that particular line can be beneficial? I would suggest to simply remove it

```suggestion
                    "of sbt-scalafix and/or Scala, or uninstalling sbt-scalafix plugin."
```

For the context, I find this confusing because
- As an sbt 1.x plugin, sbt-scalafix is only published with Scala 2.12.
- The constraint here is that for a given Scala 2.x, sbt-scalafix should use a scalafix version that uses a scalameta version for which the semanticdb compiler is published - this is very hard to check and almost equally hard to document.

bjaglin (commented):
Interestingly, that message added in https://github.com/scalacenter/sbt-scalafix/pull/369 was really to cover the combination of a recent sbt-scalafix with an old scala version, but I didn't realize some users could be on an old sbt-scalafix with a recent scala version.

Thanks for taking the time to report back and make the error message more actionable!

bjaglin:
CI doesn't run automatically for first-time contributors, I just kicked it off. Could you update https://github.com/scalacenter/sbt-scalafix/blob/72375c33a2efd7dba0f2796b4a9820874c924428/src/sbt-test/sbt-scalafix/unavailable-semanticdb-scalac/build.sbt#L22 ?

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
danicheg: Ah, yes, you're right. It can be removed.

danicheg:
@bjaglin Thank you for your assistance!

PR #469, opened 2025-02-19
scalafix 0.14.1

bjaglin:


project/Dependencies.scala
```diff
@@ -2,7 +2,7 @@ import sbt._
 
 object Dependencies {
   val x = List(1) // scalafix:ok
-  def scalafixVersion: String = "0.14.0"
+  def scalafixVersion: String = "0.14.0+51-953bdc64-SNAPSHOT"
```
bjaglin, line 5: i forgot to bump to 0.14.1 ... I'll release 0.14.2 right after since resolution of that snapshot might not work on all environments

PR #482, opened 2025-04-13
decouple sbt-scalafix & scalafix versioning

bjaglin:
Fixes https://github.com/scalacenter/scalafix/issues/1146


project/Dependencies.scala
```diff
@@ -2,11 +2,16 @@ import sbt._
...
   val x = List(1) // scalafix:ok
-  def scalafixVersion: String = "0.14.2"
+
+  // keep this as low as possible to ensure sbt-scalafix can be upgraded even if scalafix-interfaces cannot
+  def scalafixVersion: String =
+    "0.14.2+39-c2bf18de-SNAPSHOT" // https://github.com/scalacenter/scalafix/pull/2226
 
   val all = List(
+    "ch.epfl.scala" % "scalafix-interfaces" % scalafixVersion
+      exclude ("ch.epfl.scala", "scalafix-properties"),
```
bjaglin, line 12: This is the key change in this PR: `exclude` is honored in the POM, and thus will allow a sbt-scalafix-only install to load ScalafixPlugin (since classes from interfaces are present) but fail at runtime (because properties file cannot be found).

src/main/scala/scalafix/internal/sbt/ScalafixInterface.scala
```diff
@@ -151,6 +151,10 @@ object ScalafixInterface {
       callback: ScalafixMainCallback
   ): ScalafixInterface = {
 
+    // force scalafix properties to be loaded, to handle failure within sbt-scalafix instead
+    // of within scalafix-interfaces, which has a more generic error message
```
bjaglin, line 155: https://github.com/scalacenter/scalafix/blob/c2bf18de49ad00c7693e9a00ebc3c811bb72dd4d/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java#L170

project/Dependencies.scala
```diff
@@ -2,11 +2,16 @@ import sbt._
 
 object Dependencies {
   val x = List(1) // scalafix:ok
-  def scalafixVersion: String = "0.14.2"
+
+  // keep this as low as possible, to allow bumping sbt-scalafix without scalafix-interfaces
+  def scalafixInterfacesVersion: String =
+    "0.14.2+48-9b6e03ac-SNAPSHOT" // scala-steward:off
```
bjaglin, line 8: 🛑 this must be updated to a stable version before merging, or at least tagging a release

bjaglin:
@adpi2 can I get your opinion on this? I still have to write a salafix rule for a smooth steward migration in `project/plugins.sbt` and docs to update, but I am satisfied with the overall approach.

.scala-steward.conf
```diff
@@ -6,3 +6,10 @@ updates.pin = [
   # JGit 6.x requires Java 11, see https://www.eclipse.org/lists/cross-project-issues-dev/msg18654.html
   { groupId = "org.eclipse.jgit", artifactId = "org.eclipse.jgit", version = "5." },
 ]
+
+dependencyOverrides = [
+  {
+    dependency = { groupId = "ch.epfl.scala", artifactId = "scalafix-interfaces" },
+    pullRequests = { frequency = "@asap" },
```
bjaglin, line 13: this will control the `project/plugins.sbt` version

src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -89,7 +89,13 @@ object ScalafixPlugin extends AutoPlugin {
           "Defaults to a wrapper around `sbt.Logger`."
       )
 
-    val scalafixSemanticdb: ModuleID =
+    val scalafixAllowDynamicFallback: SettingKey[Boolean] =
+      settingKey[Boolean](
+        "Control whether the latest version of scalafix may be downloaded instead of failing when " +
+          "scalafix-interfaces is not explicitly provided. Off by default because non-deterministic."
+      )
```
bjaglin, line 96: this was first a private key used by scalafixEnable, but I thought we could make it public for power users, without necessarily documenting it

tgodzik (approved):
Looks good from my side. For Metals and Scala CLI we already use interfaces, so we should not be impacted, right?

In scalafmt the approach was to include the version in config to make sure we always run the correct version irrelevant of the tool used. It might be less of an issue here, since the rules are much slower to evolve and also don't usually touch the whole code as scalafmt can.

What would be much more useful is having rule deps to be defined in in the config file as it's a bit of a hassle for the user to set up for metals, but we could potentially try exporting that information also

src/sbt-test/sbt-scalafix/custom-src-directory/project/plugins.sbt
```diff
@@ -1,2 +1,4 @@
 resolvers += Resolver.sonatypeRepo("public")
 addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % sys.props("plugin.version"))
+libraryDependencies += "ch.epfl.scala" % "scalafix-interfaces" %
```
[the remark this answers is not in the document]
bjaglin: That's right, and I am planning to make that one-shot addition automatic for scala-steward users via a scalafix rule

bjaglin:
> For Metals and Scala CLI we already use interfaces, so we should not be impacted, right?

No impact indeed, scalafix-interfaces remains backward compatible for Metals, and there are no changes in scalafix-cli for scala-cli (we can't use scalafix-interfaces there because it wouldn't work on the native bundle)

> include the version in config to make sure we always run the correct version irrelevant of the tool used

> What would be much more useful is having rule deps to be defined in in the config file 

We should have both at the same time indeed (the later is tracked under https://github.com/scalacenter/scalafix/issues/1625 FWIW). And steward should know about these implicit and explicit maven coordinates in `.scalafix.conf` (just like it knows about the implicit one in `.scalafmt.conf`) to keep users up to date. Overall, it looked to me like a slightly bigger effort than keeping both as sbt dependencies like in this solution, but I agree this is the end goal which will benefit all scalafix-interfaces clients.

bjaglin:
> Overall, it looked to me like a slightly bigger effort than keeping both as steward-friendly sbt dependencies like in this solution, but I agree this is the end goal which will benefit all scalafix-interfaces clients.

Actually, since I’ve got a bit of free time coming up, I’ll give this a shot. If it turns out to be doable with limited efforts, I might just skip the intermediate step this PR was meant to be.

PR #507, opened 2025-10-01
fix scripted impacted by OSSRH shutdown

bjaglin:


src/sbt-test/sbt-scalafix/resolvers/build.sbt
```diff
@@ -1,5 +1,5 @@
 ThisBuild / scalafixDependencies := List(
-  "ch.epfl.scala" %% "example-scalafix-rule" % "4.0.0+3-e9850db5-SNAPSHOT"
+  "ch.epfl.scala" %% "example-scalafix-rule" % "6.0.0+19-2945f6b4-SNAPSHOT"
```
bjaglin, line 2: random snapshot built after
- bumping sbt-ci-release to 1.11+ to target Maven central
- migration of ch.epfl.scala to Sonatype Central completed

bjaglin:
The failure in windows was a true positive, following in https://github.com/scalacenter/sbt-scalafix/pull/508

PR #509, opened 2025-10-01
use Sonatype Central Snapshots instead of OSSRH public/snapshots

bjaglin:
Follows https://github.com/scalacenter/sbt-scalafix/pull/493


src/sbt-test/sbt-scalafix/basic/project/plugins.sbt
```diff
@@ -1,3 +1,6 @@
-resolvers += Resolver.sonatypeRepo("public")
+resolvers += MavenRepository(
+  "sonatype-central-maven-snapshots",
+  "https://central.sonatype.com/repository/maven-snapshots/"
+)
```
bjaglin, line 4: verbose explicit definition needed to support older yet tested sbt 1.x version that predate the introduction of `Resolver.sonatypeCentralSnapshots`

bjaglin:
windows failure to be fixed in https://github.com/scalacenter/sbt-scalafix/pull/508

PR #568, opened 2026-09-09
Depend on copyResources to fix a race condition

### LINKED ISSUE #2469: [Heisenbug] "Unable to load symbol table" for non-Scala tmp files in multi-module SBT projects
### Environment

- sbt-scalafix: 0.14.7
- Project: multi-module SBT build (Gatling Enterprise Cloud, maintained by [Gatling Corp](https://gatling.io), the maintainers of [Gatling OSS](https://github.com/gatling/gatling))

### Description

Since upgrading to sbt-scalafix 0.14.7 (introduced in our codebase on June 15, 2025), we intermittently observe the following error during `scalafix` execution:

```
[info] Running scalafix on 1 Scala sources
[error] error: Unable to load symbol table: /path/to/module/target/it-classes/default/commons/V00122.sql.ee95703d.tmp
```

This is a **heisenbug** — it does not occur consistently and we have not yet been able to reliably reproduce it. We will follow up with a minimal reproduction case once we find one.

### Key observations

- The problematic file is **not a Scala source file**. In the example above, it is a `.sql` file (a Flyway migration) added to the module via `unmanagedResourceDirectories` from another module.
- **It is not always the same file** that triggers the error — we have seen different `.sql.*.tmp` files across runs.
- **It is not limited to Integration Tests** — a similar error has been observed in regular (unit) tests, with a `.json.*.tmp` file.
- The file referenced in the error is a **temporary copy** (note the `.tmp` extension with a hash) that appears to be generated by the SBT build process.

### Hypothesis

It looks like scalafix (or the symbol table loader) is picking up `.tmp` files from the `target/` directories and attempting to parse them as symbol tables. This may be a regression in how the classpath or source file discovery works in 0.14.7.

### Workaround

None found yet.

### Follow-up

We are actively trying to isolate the reproduction steps. We will update this issue with a minimal reproducer if we find one.

phdoerfler:
Fixes https://github.com/scalacenter/scalafix/issues/2469, the intermittent `Unable to load symbol table: .../<file>.tmp` failures

## Cause

The semantic scalafix task builds its classpath as `dependencyClasspath :+ classDirectory` and depends on `compile`, but not on `copyResources`. scalameta walks `classDirectory` to build the symbol table (`GlobalSymbolTable` -> `ClasspathIndex.expandDirEntry` -> `Files.walkFileTree`). Since `copyResources` also writes into `classDirectory` and is not ordered before the walk, the two can run concurrently.

When a resource is copied via write-temp-then-rename, the walk can list a `<name>.<hash>.tmp`, then fail to stat it after it is renamed away, throwing `NoSuchFileException`. scalameta's visitor does not override `visitFileFailed`, so the default rethrows and aborts the walk; scalafix surfaces it as `Unable to load symbol table`.

A detailed analysis is available in https://github.com/scalacenter/scalafix/issues/2469#issuecomment-5605142943.

## Fix

Add a dependency on `config / copyResources` to the explicit-invocation branch, so resource copying completes before the classpath is walked.

## Scope and notes

- Applies only to the explicit invocation branch. The triggered branch is intentionally left dependency-free. The circular dependency comment looks scary.
- I reproduced the underlying race directly against scalameta 4.14.2: constructing `GlobalSymbolTable` over a directory with concurrent write-then-rename churn throws `NoSuchFileException` within a few iterations. I have not built an end-to-end failing project with this plugin change applied.


bjaglin (approved):
From a functional perspective, this will unnecessarily impact the throughput of scalafix executions on large sbt projects during cold starts, but I agree it's better to trade a bit of that throughput, rather than trying to fix it upstream by making the scalameta parser resilient to filesystem noise.

Thanks for the PR!

phdoerfler:
@bjaglin Thank you! I concur. And that is as someone who really dislikes the general sluggishness of large builds. But I dislike sporadic failures even more. I was really hoping "fixing" upstream would be viable and I tried to come up with arguments to support it but ultimately, it's the wrong place to change as it can not reliably determine what should happen. Not for every caller. One way out of that would be to provide some kind of error handler or error handling strategy to the code that walks the file tree. It could default to the current behavior yet allow callers like scalafix to explicitly permit skipping those transient files. Or only those that scalafix does not care about. But, well, then the sporadic behavior is back. So that's no bueno either.

The only correct **and** performant fix would be to model around the concurrent modification. Providing scalafix with an event sourced version so it could then pick up files that were added mid run and process those, too.

Simplified for this use case: a stream / iterator of files that the tree walk found. While the sbt copy task is running, it would populate that stream / iterator every time a copy-via-rename is finished. At the end of the copy task it would then signal the iterator that no more items will follow. This would then allow the consumer (scalafix) to detect that it has consumed everything there is and at the same time permit it to run concurrently.

The more I think about this, the more I am liking this idea but I fear it goes against sbt's idea of how tasks should be executed.

